PR #18, opened 2021-01-07
Tweak the issue template to match the repository

grahamc:
For example, there is no published GPG key.


t0mk:
@grahamc thanks, nobody looked at the issue template for a long time. Looks good. I'm merging.

Issue #11, opened 2020-12-15
Implement Connections resources


displague:
## What problem are you facing?
Equinix Metal offers Connections through Equinix Fabric. These are not made available through this provider.

### Implement Datasources

* [x] #41 Connections (by organization or project)
  * [x] #41 Connection Ports
* [x] #41 Virtual Circuits Datasource
* Events (We haven't implemented these for other resources. I'm not sure what benefit they provide in Terraform. Listed only for completeness.)
  * [ ] Connections (Not planned)
  * [ ] Ports (Not planned)
  * [ ] Virtual Circuits (Not planned)

### Implement Resources

* [x] Connections Resource #76
  * [x] Connection Ports (computed)
  * [x] ~Virtual Circuits (computed)~ (Use independent virtual-circuit datasource or resources)
* [x] Virtual Circuits Resource #92

Depends on ~packethost/packngo#228~ packethost/packngo#234 (more context there)

---

Initial Connection shape proposal (see comments for revisions):

```hcl
resource "metal_connection" "conn" {
  // create only
  type         = "string" // shared | dedicated
  speed        = 0        // int bps, rather than permitted "123 mbps" format
  organization = "string" // GET returns this as {href}
  metro        = "string" // returned as {href}
  facility     = "string" // returned as {href}, create either metro or facility, both will be returned

  // create and update
  name          = "string"
  description   = "string"
  contact_email = "string"
  redundancy    = "string" // primary | redundant
  tags = [
    "string"
  ]

  // read only
  id     = "3fa85f64-5717-4562-b3fc-2c963f66afa6"
  status = "string"
  ports {
    primary {
      id        = "3fa85f64-5717-4562-b3fc-2c963f66afa6"
      role      = "string"
      status    = "string"
      switch_id = "string"
      // virtual_circuits is promoted to connection level property
    }
    secondary {
      // same as primary, except "role=secondary"
    }
  }

  // set type (multiple, removable, updateable)
  // these are read from "ports" above, included in /connection/{id}/ports/{id}
  // (virtual_circuits list can come from either port, they should have the same list,
  // we can map by id when sync'ing, just-in-case)
  // these are updated or deleted through /virtual-circuits/{id}
  virtual_circuit {
    // create only
    nni_vlan = 0
    project = "" // returned as {href}

    // create and update
    name        = "string"
    description = "string"
    speed       = 0
    tags = [
      "string"
    ]
    vnid = 0

    // read only
    status          = "string"
[...]

PR #21, opened 2021-02-01
bump packngo dep to 0.6.0

### LINKED ISSUE #22: RetryMax and RetryWaitMax constants should be promoted to provider arguments
The RetryMax and RetryWaitMax constants should be promoted to provider arguments and/or environment variables rather than being hardcoded [here](https://github.com/equinix/terraform-provider-metal/blob/main/metal/config.go#L54-L55).

RetryMax -> max_retries? https://registry.terraform.io/providers/hashicorp/aws/latest/docs#max_retries
RetryWaitMax -> max_retry_wait?

_Originally posted by @displague in https://github.com/equinix/terraform-provider-metal/pull/21#discussion_r568028267_

t0mk:
bump packngo dep to 0.6.0 to bring the Connections support

Part of #11

Fixes #22


t0mk:
.. I wonder how the acceptance tests will fly

t0mk:
.. now it builds. I'm waiting on how the acceptance test report will look like. @displague feel free to check the test report too when it's done.

metal/config.go
```diff
@@ -49,14 +87,7 @@ func MetalRetryPolicy(ctx context.Context, resp *http.Response, err error) (bool
...
-	httpClient.RetryWaitMin = time.Second
-	httpClient.RetryWaitMax = 30 * time.Second
-	httpClient.RetryMax = 10
-	httpClient.CheckRetry = MetalRetryPolicy
-	httpClient.HTTPClient.Transport = logging.NewTransport(
-		"Equinix Metal",
-		httpClient.HTTPClient.Transport)
-
+	transport := logging.NewTransport("Equinix Metal", http.DefaultTransport)
+	httpClient := &http.Client{Transport: newRetryableTransport(transport)}
```
displague, line 91: Is there any advantage of these client changes over what was proposed in https://github.com/packethost/packngo/pull/222#issue-533950813 (with less change)?

displague:
```
retryClient := retryablehttp.NewClient()
retryClient.RetryMax = 10
retryClient.RetryWaitMin = time.Second
retryClient.RetryWaitMax = 30 * time.Second
retryClient.RetryMax = 10
retryClient.CheckRetry = packngo.RetryPolicy
httpClient := retryClient.StandardClient()
c := NewClientWithAuth(consumerToken, apiToken, httpClient)
```

displague: I think it should be `httpClient.HTTPClient.Transport`.

https://github.com/hashicorp/go-retryablehttp/blob/991b9d0a42d13014e3689dd49a94c02be01f4237/client.go#L772

displague: As a separate issue, the RetryMax and RetryWaitMax constants should be promoted to provider arguments and/or environment variables rather than being hardcoded here.

RetryMax -> max_retries? https://registry.terraform.io/providers/hashicorp/aws/latest/docs#max_retries
RetryWaitMax -> max_retry_wait?

t0mk: @displague thanks for the review, yeah you're right, your code is simpler and works. I fixed it.

I also added the retry params to the provider.

t0mk: .. but I think that pulling the retry policy from packngo is not a good idea, the provider should define its own (just like any other tool using packngo as a dep).

t0mk:
There were 2 interesting errors from the first acc test run:

One from L2 Hybrid basic test:
```
 --- FAIL: TestAccMetalPortVlanAttachment_HybridBasic (319.13s)
    testing.go:569: Step 1 error: errors during apply:
        
        Error: Provider produced inconsistent result after apply
        
        When applying changes to metal_port_vlan_attachment.test, provider "metal"
        produced an unexpected new value for was present, but now absent.
        
        This is a bug in the provider, which should be reported in the provider's own
        issue tracker.
        
```

and SSH Key projectImportBasic: 

```
--- FAIL: TestAccMetalSSHKey_projectImportBasic (2.27s)
    testing.go:569: Step 1 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.
        
        (map[string]string) {
        }
        
        
        (map[string]string) (len=1) {
         (string) (len=10) "project_id": (string) (len=36) "96e8fa35-58df-487b-be70-63d05f7a9ea7"
        }
        
=== RUN   TestAccMetalSSHKey_importBasic
```

displague:
I created #23 and #24 for the test failures.

metal/config.go
displague: The old PR comment suggested the packngo RetryPolicy. The code snippet above continues to use the MetalRetryPolicy.

metal/config.go
```diff
@@ -49,14 +51,13 @@ func MetalRetryPolicy(ctx context.Context, resp *http.Response, err error) (bool
...
-		"Equinix Metal",
-		httpClient.HTTPClient.Transport)
-
+	transport := logging.NewTransport("Equinix Metal", http.DefaultTransport)
+	retryClient := retryablehttp.NewClient()
+	retryClient.HTTPClient.Transport = transport
+	retryClient.RetryMax = c.MaxRetries
+	retryClient.RetryWaitMin = time.Second
+	retryClient.RetryWaitMax = c.MaxRetryWait
+	retryClient.CheckRetry = packngo.RetryPolicy
```
displague, line 60: Let's continue to use `MetalRetryPolicy` (unless it is identical to the `packngo` copy, then perhaps we can remove the duplication here).

metal/config.go
displague: (You may be seeing an older copy of my edited comment)

metal/config.go
```diff
@@ -49,14 +51,13 @@ func MetalRetryPolicy(ctx context.Context, resp *http.Response, err error) (bool
...
-	httpClient.RetryWaitMax = 30 * time.Second
-	httpClient.RetryMax = 10
-	httpClient.CheckRetry = MetalRetryPolicy
-	httpClient.HTTPClient.Transport = logging.NewTransport(
-		"Equinix Metal",
-		httpClient.HTTPClient.Transport)
-
+	transport := logging.NewTransport("Equinix Metal", http.DefaultTransport)
+	retryClient := retryablehttp.NewClient()
+	retryClient.HTTPClient.Transport = transport
```
displague, line 56: I don't think this will work because of what `.StandardClient()` does:
https://github.com/hashicorp/go-retryablehttp/blob/991b9d0a42d13014e3689dd49a94c02be01f4237/client.go#L772

What do you think about this: https://github.com/equinix/terraform-provider-metal/pull/21#discussion_r568004813

metal/config.go
t0mk: The policy in pakcngo and the policy in the provider are the same. I would use the policy funciton from the provider. The policy function in packngo can be removed?

metal/config.go
t0mk: It works, the RoundTripper wraps the retryable client which has it's own http.Client which still keeps the transport:
```
	transport := logging.NewTransport("Equinix Metal", http.DefaultTransport)
	retryClient := retryablehttp.NewClient()
	retryClient.HTTPClient.Transport = transport
        [...]
	httpClient := retryClient.StandardClient()
	t := httpClient.Transport
	rhrt := t.(*retryablehttp.RoundTripper)
	fmt.Println("EEEEEEEEEEEEEEEEEE", rhrt.Client.HTTPClient.Transport)
```

.. prints `EEEEEEEEEEEEEEEEEE &{Equinix Metal 0x1b2f4e0}`

> What do you think about this: #21 (comment)
> `httpClient := retryClient.StandardClient()`
> `httpClient.HTTPClient.Transport = transport`

That wouldn't compile.

displague:
More interesting errors that seem unrelated.  I wonder if the two versions of TF acceptance tests are running side-by-side and the sweeper from one test is knocking out the test resources of the other test.  Most of these errors could be explained by a resource being deleted by an external actor.

```
--- FAIL: TestAccMetalDataSourceProject_Basic (2.42s)
    testing.go:569: Step 0 error: errors during apply:
        
        Error: Not found
        
          on /tmp/tf-test475229150/main.tf line 6:
          (source code not available)
```

```
--- FAIL: TestAccMetalProject_BGPBasic (0.77s)
    testing.go:569: Step 0 error: errors during apply:
        
        Error: Not found
        
          on /tmp/tf-test218057823/main.tf line 2:
          (source code not available)
```

```
--- FAIL: TestAccMetalProject_BGPUpdate (1.39s)
    testing.go:569: Step 1 error: errors during apply:
        
        Error: Not found
        
          on /tmp/tf-test871845553/main.tf line 2:
          (source code not available)
```


```
--- FAIL: TestAccMetalPortVlanAttachment_L2Individual (1017.40s)
    testing.go:569: Step 1 error: errors during apply:
        
        Error: POST https://api.equinix.com/metal/v1/ports/4779f6fa-e488-47fb-9ca1-b5c683224e1b/convert/layer-2: 403 You are not authorized to view this resource 
        
          on /tmp/tf-test315818338/main.tf line 29:
          (source code not available)
```

displague:
@t0mk do you think https://github.com/equinix/terraform-provider-metal/issues/23#issuecomment-772577057 is evidence that the retry policy was not used? 

https://github.com/equinix/terraform-provider-metal/blob/main/metal/config.go#L26-L48

metal/config.go
displague, line 60: Maybe we should just use the default from retryablehttp? It seems to do everything we are doing and more. Do you happen to know why we use a custom policy?

https://github.com/hashicorp/go-retryablehttp/blob/master/client.go#L399-L463

displague:
I'll merge this as-is and we can revisit the retry / 504 issue as part of #23. I would like to get that resolved before the next release, but 5xx handling can't be guaranteed IMHO, so it's ok if we slip on that.

PR #29, opened 2021-02-10
introduce METAL_AUTH_TOKEN to replace PACKET_AUTH_TOKEN

### LINKED ISSUE #26: Introduce METAL_AUTH_TOKEN environment variable
Introduce `auth_token` environment variable `METAL_AUTH_TOKEN` with higher priority than `PACKET_AUTH_TOKEN`

https://github.com/equinix/terraform-provider-metal/pull/21/files#diff-f83e6f29b5ad609210e057442d576d3c8ff966ea294706f37c89d102fd3059f1R20

There should also be documentation updates related to this change.

We should also deprecate the `PACKET_AUTH_TOKEN`. It should be trivial to continue to support the legacy environment variable for some period of time.

displague:
`METAL_AUTH_TOKEN` will be used if set, otherwise `PACKET_AUTH_TOKEN` will
continue to be used.

`METAL_AUTH_TOKEN` has not been added to the GitHub Action secrets. The action is still pulling from the old variable name, for now. 

Fixes #26

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -66,7 +66,8 @@ resource "metal_device" "web1" {
 The following arguments are supported:
 
 * `auth_token` - (Required) This is your Equinix Metal API Auth token. This can
-  also be specified with the `METAL_AUTH_TOKEN` or `PACKET_AUTH_TOKEN` shell
-  environment variable.
+  also be specified with the `METAL_AUTH_TOKEN` environment variable.
+
+  Use of the legacy `PACKET_AUTH_TOKEN` environment variable is deprecated.
 * `max_retries` - Maximum number of retries in case of network failure.
 * `max_retry_wait_seconds` - Maximum time to wait in case of network failure.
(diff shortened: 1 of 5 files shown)


docs/index.md
```diff
@@ -65,7 +65,8 @@ resource "metal_device" "web1" {
 
 The following arguments are supported:
 
-* `auth_token` - (Required) This is your Equinix Metal API Auth token. This can also be specified
-  with the `PACKET_AUTH_TOKEN` shell environment variable.
+* `auth_token` - (Required) This is your Equinix Metal API Auth token. This can
```
t0mk, line 68: maybe don't mention PACKET_AUTH_TOKEN since we deprecate it

t0mk:
@displague looks good

docs/index.md
displague: I'll still mention it because I don't want people to be surprised when things still work because they forgot they have `PACKET_AUTH_TOKEN` set. I'll mention that it is deprecated.

displague: 968aad2 includes this message

PR #30, opened 2021-02-10
use Terraform and provider version as packngo UserAgent

displague:
Fixes #19 

When tested locally using `TF_LOG=DEBUG TF_ACC=1 go test ./... -v -timeout=20m -run=TestAcc...`:

```
User-Agent: HashiCorp Terraform/0.11+compatible (+https://www.terraform.io) Terraform Plugin SDK/1.0.0 terraform-provider-metal/dev
```

Issue #28, opened 2021-02-05
Support for Metros


displague:
This issue is a planning/placeholder for incoming work related to backend work in progress:

https://feedback.equinixmetal.com/platform/p/introduce-locations-concept-including-support-for-metros

As information becomes available this description will be updated and comments will be added to support the implementation changes.

Depends on https://github.com/packethost/packngo/issues/238

We wil need a new datasource:
* [x] datasource `metal_metro` #58

We will also need new metro properties. Generally, these will be optional and adjacent to facility. Facility will become optional, when metro is specified. The metro property will be a slug. Users can fetch additional metro details through a lookup on the `metal_metro` datasource.

The following need a read-only metro property:
* [ ] datasource metal_spot_market_request _(Update: request_id is the only parameter, and device_ids the only attribute)_
* [x] datasource metal_spot_market_price metro parameter (make facility optional) #64
* [x] datasource metal_ip_block_ranges metro filter #58
* [x] datasource metal_precreated_ip_block metro property #58
* [ ] datasource metal_volume metro property _(Update: volumes can not be metro scoped)_
* [x] datasource metal_facility metro property #31 #58

The following need an optional metro property as an alternative to the facility. One is required, the values should not conflict, but if they do the API response will handle that - we will not try to prevalidate.

* [x] resource metal_device metro property #46 
* [x] resource metal_spot_market_request  #63 
* [x] resource `metal_vlan` #53 
* [x] resource `reserved_ip_block` #56 

Examples should be broadly be updated to demonstrate metros. A few smaller examples can demonstrate Facilities.

* [x] Examples are updated to use Metros #65
* [x] #36 Examples are updated to use Metro supported plans and images #65

displague:
We wil need a:
* datasource metal_metro
* datasource metal_spot_market_request metro property
* datasource metal_ip_block_ranges metro property
* datasource metal_spot_market_price metro property
* datasource metal_precreated_ip_block metro property
* datasource metal_volume metro property

* resource metal_device metro property (optional and adjacent to facility, behaves in the same way. One of metro of facility is required)
* resource metal_spot_market_request metro property (optional and adjacent to facility, behaves in the same way. One of metro of facility is required)

* resource metal_vlan metro property (optional and adjacent to facility, behaves in the same way. One of metro of facility is required)
* resource reserved_ip_block (optional and adjacent to facility, behaves in the same way. One of metro of facility is required)

displague:
[...]

PR #31, opened 2021-02-10
Datasource metal_facility

displague:
I needed a simple resource to test #30 so I added this non-sensitive data resource that can be tested quickly without creating remote resources.

We may wish to add the ability to lookup facility by ID in the future, I found that all of the existing Terraform resources only expose facility code, so this seemed like the best attribute to start with.

An auth token environment variable is still needed to run the test, there are no unauthenticated endpoints in the EM API.

```
TF_LOG=DEBUG TF_ACC=1 go test ./... -v -timeout=20m -run=TestAccFacilityDataSource_Basic
```

### DIFF
--- a/docs/data-sources/facility.md
+++ b/docs/data-sources/facility.md
@@ -1,0 +1,42 @@
+---
+page_title: "Equinix Metal: metal_facility"
+subcategory: ""
+description: |-
+  Provides an Equinix Metal facility datasource. This can be used to read facilities.
+---
+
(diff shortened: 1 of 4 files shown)


displague:
Created https://github.com/packethost/packngo/issues/243 from an examination of the facility API debug output.

displague:
I don't see any reason to rush on this. We can pick this up with #28 or merge this in advance of that related but independent metros work.

PR #46, opened 2021-03-25
Metros support in metal_device

t0mk:
This is a canary PR to enable metros in metal_device.

Depends on packethost/packngo#246 (not yet tagged)


displague:
@t0mk when either metros or facilities can be set and either can be computed, we'll want to make sure that the Terraform provider doesn't treat this as a tainted condition, since changing either a metro or a facility would otherwise require the device to be recreated. Likewise, the computed values may provide both a metro and a facility, in this case the provider should not be concerned that both values exist.

The separate Facility and DeployedFacility fields help us with this, but Metro does not need a DeployedMetro field because it does not support arrays of potential metros, nor a special "any" metro (features that facilities offer).  (_I'll verify that metros do not support these features._)

You may have already addressed those concerns, I'm just surfacing the potential problems.

displague:
> _I'll verify that metros do not support these features_

Metro based VLANs will only report a metro.
Devices will always report facility, and only metro if deployed to a metro.

displague:
Legacy VLANs show facility, not metro, and can not be attached outside of the facility.
Legacy IP Reservations show facility, not metro,  and can not be attached outside of the facility.
Legacy Connections show facility, not metro,  and can not be attached outside of the facility.

Generally:
* if the facility is unchanged and the metro has changed - the diff should be suppressed.
* if the facility is changed to unset and the metro has been set - the diff should be suppressed.

t0mk:
@displague OK, this all sounds really messy, but I will try to pay attention to it, and test the provider upgrade.

t0mk:
- [x] device created with metal prov 1.1.0 in a non-metro fac. Update to build from this PR -> `terraform plan` is clean
- [ ] device created with metal prov 1.1.0 in a metro fac. Update to build from this PR -> `terraform plan` is clean

t0mk:
@displague thanks for listing the migration use cases, but tet's keep this PR only for metal_device.

I listed 2 scenarios that I tested in the previous comment. Add some if you'd like.

Note that it's hard to test how a legacy-API-created device will be treated by the metro-enabled TF provider. The org where I test is already metro-enabled. I can't create a device in current prod API environment, switch ti metro-enabled API and check `tf plan`. I hope I explained it well.

This PR is ready for review. I will start preparing PRs for vlans and IP reservations.

metal/resource_metal_device.go
```diff
@@ -422,7 +454,7 @@ func resourceMetalDeviceCreate(d *schema.ResourceData, meta interface{}) error {
 func resourceMetalDeviceRead(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*packngo.Client)
 
-	device, _, err := client.Devices.Get(d.Id(), &packngo.GetOptions{Includes: []string{"project"}})
+	device, _, err := client.Devices.Get(d.Id(), &packngo.GetOptions{Includes: []string{"project", "metro"}})
```
displague, line 457: I wonder if we should be defensive and include `facility` too, just incase the default inclusion of facility changes in the future.

t0mk: good idea. done in 8e5bc05

displague:
If a user starts with

```hcl
resource "metal_device" "foo" {
  facility = "any"
}
```

and refreshes to:

```hcl
resource "metal_device" "foo" {
  facility = "any"
  # in state:
  # deployed_facility = "ewr1"
  # metro = "ny"
}
```

And updates their config to:
```hcl
resource "metal_device" "foo" {
  metro = "ny"
  # in state:
  # deployed_facility = "ewr1"
  # metro = "ny"
  
  # changed: facility is now empty, was "any"
}
```

Would the diff be suppressed?
I think this is somewhat of an edge-case since the user would have had to actively change both their stated facility and stated metro.

Does this PR introduce any breaking changes that would make the next release 2.0.0 rather than 1.2.0?

displague:
@t0mk I've created https://github.com/equinix/terraform-provider-metal/discussions/55 as a place for ongoing discussion about edge-cases and expected behavior.

t0mk:
@displague I will try that scenario tomorrow.

t0mk:
@displague I updated go to 1.16 adn build succeeds. If you OK, I will merge, and then rebase the `metro` PRs.

displague:
I'm merging this into master so we can have some time with the full set of changes. I think this is generally ready and we may have some edge cases to document, but we are generally in good shape with what we are testing for.

PR #53, opened 2021-04-02
Add "metro" parameter to metal_vlan resource

t0mk:
Addresses #28 

`metro` param conflicts with `facility`

I will still need to check migration of vlans created with old provider version.

### DIFF
--- a/metal/resource_metal_vlan.go
+++ b/metal/resource_metal_vlan.go
@@ -32,6 +32,7 @@ func resourceMetalVlan() *schema.Resource {
 				Optional:      true,
 				ForceNew:      true,
 				ConflictsWith: []string{"metro"},
+				Deprecated:    "Use metro instead",
 			},
 			"metro": {
 				Type:          schema.TypeString,
--- a/metal/resource_metal_vlan.go
+++ b/metal/resource_metal_vlan.go
@@ -1,6 +1,8 @@
 package metal
 
 import (
+	"errors"
+
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/packethost/packngo"
 )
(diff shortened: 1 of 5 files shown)


metal/resource_metal_vlan.go
```diff
@@ -40,10 +50,23 @@ func resourceMetalVlan() *schema.Resource {
...
+
+	if !facOk && !metroOk {
+		return friendlyError(errors.New("one of facility and metro must be configured"))
+	}
+
 	createRequest := &packngo.VirtualNetworkCreateRequest{
 		ProjectID:   d.Get("project_id").(string),
 		Description: d.Get("description").(string),
-		Facility:    d.Get("facility").(string),
+	}
```
displague, line 64: Let's add the VXLAN request field (which will be getting renamed from VLAN to VXLAN soon/today) https://github.com/packethost/packngo/pull/257

metal/resource_metal_vlan.go
```diff
@@ -26,9 +28,17 @@ func resourceMetalVlan() *schema.Resource {
...
 			},
 			"facility": {
-				Type:     schema.TypeString,
-				Required: true,
-				ForceNew: true,
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"metro"},
+				Deprecated:    "Use metro instead",
```
displague, line 35: Facilities are not deprecated.

metal/resource_metal_vlan.go
displague: It can only be used with Metro.

displague:
> I will still need to check migration of vlans created with old provider version.

Old VLANs created before metros was an option will include facility only.
New VLANs created in a facility will act the same way.

New VLANs created in a metro will return both a facility and a metro, so we will need to make sure this is handled correctly (not getting the user stuck in a conflict situation).  Maybe we should introduce a `deployed_facility` field so that read-only facilities from Metro based provisioning is not confused with the facility request field.  This would be consistent with the device resource, but it does introduce some discrepancies from the API spec.  I think it's tolerable because of how Terraform resource computed/optional/required fields work in practice.

t0mk:
> New VLANs, created in a metro, will return both a facility and a metro.

The newly created VLAN in metro doesn't return facility.

I made it so that if metro is set from "" to some string, and facility is set (in TF state), the diff is suppressed.

metal/resource_metal_vlan.go
displague: https://github.com/packethost/packngo/releases/tag/v0.10.0

t0mk:
@displague I fixed your review comments. Please take another look when you have time.

t0mk:
rebased and ready for review

PR #56, opened 2021-04-08
metro field in metal_reserved_ip_block

t0mk:
This PR adds `metro` param to reserved IP block. It also bumps packngo dep to 0.10.0.

#28 
#55

### DIFF
--- a/metal/resource_metal_reserved_ip_block.go
+++ b/metal/resource_metal_reserved_ip_block.go
@@ -17,24 +17,28 @@ func metalIPComputedFields() map[string]*schema.Schema {
 			Computed: true,
 		},
 		"address_family": {
-			Type:     schema.TypeInt,
-			Computed: true,
+			Type:        schema.TypeInt,
+			Computed:    true,
+			Description: "Address family as integer (4 or 6)",
 		},
 		"cidr": {
-			Type:     schema.TypeInt,
-			Computed: true,
+			Type:        schema.TypeInt,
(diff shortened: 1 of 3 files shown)


t0mk:
@displague 

I saw this is Github actions build log:
```
Error: ../../../go/pkg/mod/github.com/packethost/packngo@v0.10.0/packngo.go:257:22: resp.Header.Values undefined (type http.Header has no field or method Values)
note: module requires Go 1.15
```

Is that a problem?

displague:
> Is that a problem?

I don't think it should be a problem to update the go build version. Let's find out in the GH action and update the versions we build with throughout.

https://github.com/hashicorp/terraform-provider-google-beta/blob/master/go.mod#L36

metal/resource_metal_reserved_ip_block.go
```diff
@@ -72,9 +72,16 @@ func resourceMetalReservedIPBlock() *schema.Resource {
 		ForceNew: true,
 	}
 	reservedBlockSchema["facility"] = &schema.Schema{
-		Type:     schema.TypeString,
-		Optional: true,
-		ForceNew: true,
+		Type:          schema.TypeString,
+		Optional:      true,
+		ForceNew:      true,
+		ConflictsWith: []string{"metro"},
```
displague, line 78: Let's get descriptions added while we are here.

t0mk:
@displague I fixed your review comments

PR #57, opened 2021-04-08
update packngo to 0.10.0 and golang to 1.15

t0mk:

### DIFF
--- a/.github/workflows/acctest.yml
+++ b/.github/workflows/acctest.yml
@@ -17,7 +17,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v2.1.3
       with:
-        go-version: '1.13'
+        go-version: '1.15'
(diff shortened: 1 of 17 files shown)


t0mk:
:crossed_fingers: :crossed_fingers: :crossed_fingers:  for the integration tests in github actions

displague:
https://github.com/hashicorp/terraform-provider-google-beta/blob/master/go.mod#L36 - It seems we should be able to adopt 1.16 while we are at it.

t0mk:
I updated golang to 1.16 :crossed_fingers:

t0mk:
@displague I updated go to 1.16 adn build succeeds. If you OK, I will merge, and then rebase the metro PRs.

displague:
The acceptance tests were almost 100%!

PR #58, opened 2021-04-09
Metro datasource updates

displague:
Implements components of #28 

* Adds `metal_metro` datasource
* Adds `metro` to `metal_facility` datasource
* Adds `metro` to `metal_precreate_ip_block` datasource
* Adds `metro` as an optional search parameter for `metal_ip_block_ranges`

### DIFF
--- a/metal/datasource_metal_metro.go
+++ b/metal/datasource_metal_metro.go
@@ -14,7 +14,6 @@ func dataSourceMetalMetro() *schema.Resource {
 			"id": {
 				Type:        schema.TypeString,
 				Description: "The ID of this Metro.",
-				Optional:    true,
 				Computed:    true,
 			},
 			"code": {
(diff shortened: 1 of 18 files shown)


metal/datasource_metal_metro.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/packethost/packngo"
+)
+
+func dataSourceMetalMetro() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceMetalMetroRead,
+		Schema: map[string]*schema.Schema{
+			"id": {
```
t0mk, line 14: `id`, `country` and `name` are not input arguments - having "Optional" set is useless here

There are similar issues with fields in the facility datasource, probably my fault. Can you fix those too please?

t0mk (approved):

PR #52, opened 2021-04-02
Fix "make errcheck" and golangci-lint warnings

displague:
This is a port of https://github.com/packethost/terraform-provider-packet/pull/301.

Because this PR surface errors that were previously ignored, any surfaced errors should be fixed (not just handled properly) before merging.

PR #63, opened 2021-04-12
Adds metros to SpotMarketRequest resources

displague:
Part of #28

### DIFF
--- a/metal/datasource_metal_spot_market_request_test.go
+++ b/metal/datasource_metal_spot_market_request_test.go
@@ -57,10 +57,10 @@ resource "metal_project" "test" {
 
 resource "metal_spot_market_request" "req" {
   project_id    = "${metal_project.test.id}"
-  max_bid_price = 0.2
+  max_bid_price = 0.01
   facilities    = ["sjc1"]
-  devices_min   = 2
-  devices_max   = 2
+  devices_min   = 1
+  devices_max   = 1
   wait_for_devices = false
 
   instance_parameters {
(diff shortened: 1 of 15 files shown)


metal/datasource_metal_spot_market_request_test.go
```diff
@@ -21,9 +24,24 @@ func TestAccDataSourceMetalSpotMarketRequest_Basic(t *testing.T) {
 			{
 				Config: testDataSourceMetalSpotMarketRequestConfig_Basic(projectName),
 				Check: resource.ComposeTestCheckFunc(
-					testAccCheckMetalSpotMarketRequestExists("metal_spot_market_request.req", &key),
-					resource.TestCheckResourceAttr(
-						"data.metal_spot_market_request.dreq", "device_ids.#", "2"),
```
displague, line 26: I modified this test to not wait for devices and to offer so low a bid that it should not be accepted (to make sure the SMR can be quickly created and deleted).

The PlanOnly step ensures that a change from a facility to a Metro based SMR triggers a taint. The ID comparison in the following step ensures a new SMR was created.

I moved the data resource device count check to the resource_spot_market_request check, which is still configured to "wait" for the devices.

metal/errors.go
```diff
@@ -41,13 +41,19 @@ func friendlyError(err error) error {
 }
 
 func isForbidden(err error) bool {
+	if r, ok := err.(*packngo.ErrorResponse); ok && r.Response != nil {
```
displague, line 44: Cherry-picked from #52

PR #65, opened 2021-04-13
remove Packet legacy from docs (facilities, plans)

### LINKED ISSUE #36: Documentation (examples) should use Equinix Metal plans and facilities
The documentation includes examples that will not work for many new users to Equinix Metal because the plans and facilities use legacy Packet API slugs.

Update the documentation to use modern slugs.

t0mk:
fixes #36 

### DIFF
--- a/docs/resources/project_ssh_key.md
+++ b/docs/resources/project_ssh_key.md
@@ -25,7 +25,7 @@ resource "metal_project_ssh_key" "test" {
 
 resource "metal_device" "test" {
   hostname            = "test"
-  plan                = "baremetal_0"
+  plan                = "c3.medium.x86"
   facilities          = ["ny5"]
   operating_system    = "ubuntu_20_04"
   billing_cycle       = "hourly"
--- a/docs/resources/project_ssh_key.md
+++ b/docs/resources/project_ssh_key.md
@@ -26,8 +26,8 @@ resource "metal_project_ssh_key" "test" {
 resource "metal_device" "test" {
   hostname            = "test"
   plan                = "baremetal_0"
-  facilities          = ["ewr1"]
-  operating_system    = "ubuntu_16_04"
+  facilities          = ["ny5"]
+  operating_system    = "ubuntu_20_04"
   billing_cycle       = "hourly"
   project_ssh_key_ids = [metal_project_ssh_key.test.id]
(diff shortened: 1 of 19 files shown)


t0mk:
@displague Please take a look and let me know if you spot more legacy values. If OK, feel free to merge.

docs/resources/project_ssh_key.md
```diff
@@ -26,8 +26,8 @@ resource "metal_project_ssh_key" "test" {
 resource "metal_device" "test" {
   hostname            = "test"
   plan                = "baremetal_0"
```
displague, line 28: `git grep plan.*[12_] metal/ docs/ | grep -v _test.go` there are a few legacy plans captured in this grep.

displague:
On the provider page example (https://registry.terraform.io/providers/equinix/metal/latest/docs#example-usage), I think we should include a metal_project instead of the `locals` project_id.  I think this is more comprehensive and less complex at the same time.  We should use metro instead of facility in the first example too.

We should update most of the examples on https://registry.terraform.io/providers/equinix/metal/latest/docs/resources/device#example-usage to use metros too.  I think we can leave one or two facility/facilities examples.

t0mk:
@displague 

> On the provider page example (https://registry.terraform.io/providers/equinix/metal/latest/docs#example-usage), I think we should include a metal_project instead of the locals project_id. 

I changed it the other way around couple of years ago beause an exec from Packet asked me to. When people tried the exapmles and had a lot of projects created, they would get monthly bills per project and it was confustion. I will try to find the issue.

docs/resources/project_ssh_key.md
t0mk: thx will fix later today

t0mk:
> We should update most of the examples on https://registry.terraform.io/providers/equinix/metal/latest/docs/resources/device#example-usage to use metros too. I think we can leave one or two facility/facilities examples.

OK, will do later today.

displague:
> When people tried the examples and had a lot of projects created, they would get monthly bills per project and it was confusion.

That's unfortunate for clear documentation, but makes sense :-)
Maybe we could use a project data source to meet both goals?

```hcl
data "metal_project" "project" {
  name = "My Project"
}
```

PR #66, opened 2021-04-13
specify test location dynamically in L2 VLAN attachment tests

displague:
Formerly https://github.com/packethost/terraform-provider-packet/pull/215

Signed-off-by: Marques Johansson <mjohansson@equinix.com>

### DIFF
--- a/metal/resource_metal_port_vlan_attachment_test.go
+++ b/metal/resource_metal_port_vlan_attachment_test.go
@@ -21,10 +21,10 @@ resource "metal_project" "test" {
 resource "metal_device" "test" {
   hostname         = "tfacc-device-port-vlan-attachment-test"
   plan             = "s1.large.x86"
-  facilities       = ["nrt1"]
+  metro            = "da"
   operating_system = "ubuntu_16_04"
   billing_cycle    = "hourly"
-  project_id       = "${metal_project.test.id}"
+  project_id       = metal_project.test.id
 }
 `, name)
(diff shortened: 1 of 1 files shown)


displague:
@t0mk What do you think of making this change? I could see keeping some `facility` tests around, or having mixed facility vlan  + metro device tests, but in this PR I was just hoping to close out one of our old PRs on the packet provider and this seemed like the most straight-forward way to do that.

metal/resource_metal_port_vlan_attachment_test.go
```diff
@@ -212,12 +212,12 @@ resource "metal_device_network_type" "test" {
 resource "metal_vlan" "test" {
   description = "test vlan"
   facility    = "dfw2"
```
t0mk, line 214: maybe we can subst the facility string literal here with `metal_device.test.deployed_facility`

t0mk:
I see what you tried to do and it makes sense.

I tried to run test `TestAccMetalPortVlanAttachment_L2Bonded` and got 

```
POST https://api.equinix.com/metal/v1/ports/17e6b0bb-0a0b-4247-95a0-54edaff68ec9/convert/layer-2:
   504  Unexpected Content-Type text/html with status 504 Gateway Time-out
```

.. but that's probably unrelated.

displague:
The 504 error occurred in the GH Action test too, something is not right here.

PR #67, opened 2021-04-14
metal_vlan datasource

### LINKED ISSUE #51: Add a Data Source for VLANs
It would be helpful to have data source for VLANs after they get created. Currently the Terraform provider only has a `VXLAN` attribute for the VLAN resource.

t0mk:
This PR adds a `metal_vlan` datasource. 

I didn't test the `assigned_devices_ids`.

@displague please review. I did this in a bit of haste, feel free to comment and fix.

fixes #51


metal/datasource_metal_vlan.go
```diff
@@ -0,0 +1,124 @@
...
+				Computed:      true,
+				ConflictsWith: []string{"vlan_id"},
+				Description:   "ID of parent project of the VLAN. Use together with vxland",
+			},
+			"vxlan": {
+				Type:          schema.TypeInt,
+				Optional:      true,
+				Computed:      true,
+				ConflictsWith: []string{"vlan_id"},
+				Description:   "VXLAN numner of the VLAN. Unique in a project. Use with project_id",
```
displague, line 29: https://github.com/equinix/terraform-provider-metal/issues/69#issuecomment-820445819

It sounds like we can't count on these being unique.

How does this sound?

```hcl
data "metal_vlan" "vxlan1000" {
 vxlan = 1000
 project_id = "required"
 
 metro = "optional"
 facility = "optional"
 
 // error if more than 1 match
}
```

t0mk: I've just came here to point this out. I will need to rework the lookup.

displague: I asked about convention for this kind of optional filtering here: https://hashicorp.slack.com/archives/C0HSE5WMB/p1618495823010700

I can't think of examples of this, but I expect it will be fine.

displague: Seems we do this here: https://github.com/equinix/terraform-provider-metal/blob/main/metal/datasource_metal_project.go#L165

t0mk:
@displague I reworked the lookup. Now user must also pass facility or metro. Now the result should be unique by the API constraints. Otherwise it will pick the first matching vlan.

docs/data-sources/vlan.md
```diff
@@ -0,0 +1,60 @@
...
+        metro      = "sv"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `vlan_id` - Metal UUID of the VLAN resource to look up
+* `project_id` - UUID of parent project of the VLAN. Use together with the vxlan number and metro or facility`
```
displague, line 49: trailing backtick here on project_id.

I think we generally say optional or required (or both?)  when listing the arguments

docs/data-sources/vlan.md
```diff
@@ -0,0 +1,60 @@
...
+* `vxlan` - vxlan number of the VLAN to look up. Use together with the project_id and metro or facility`
```
displague, line 50: backtick here too

metal/datasource_metal_vlan.go
```diff
@@ -0,0 +1,138 @@
...
+		vlans, _, err := c.ProjectVirtualNetworks.List(
+			projectRaw.(string),
+			&packngo.GetOptions{Includes: []string{"assigned_to"}},
+		)
+		if err != nil {
+			return friendlyError(err)
+		}
+		for _, v := range vlans.VirtualNetworks {
+			if v.VXLAN == vxlan {
+				if (metroOk && (metro == v.MetroCode)) ||
```
displague, line 106: This logic makes metro or facility required.

We can make facility, metro, and even vxlan optional for users that have only one vlan in a project:

```go
matches := 0
for _, v := range vlans.VirtualNetworks {
  if vxlan != "" && v.VXLAN != vxlan {
    continue
  }
  if facility != "" && v.FacilityCode != facility {
    continue
  }
  if metro != "" && v.MetroCode != metro {
    continue
  }
  matches++
  if matches > 1 { return err }
  vlan = &v 
}

if matches == 0 { return err }
```

docs/data-sources/vlan.md
displague: Everything is optional, I was thinking project_id was required, but not in the presence of vlan_id. I suppose this is the usual format then. Perhaps we could add that Either vlan_id or project_id is required?

metal/datasource_metal_vlan.go
displague: I've added these changes and an added test to verify.

docs/data-sources/vlan.md
displague: fixed

PR #80, opened 2021-04-21
Add datasource metal_reserved_ip_block, docs and a test

### LINKED ISSUE #68: no way to lookup ip reservation ids
When consuming reservation ids, in a device's ip_address block, the reservation id is a needed field.

It is not currently possible to get the reservation id of an existing reservation.

Perhaps ip_block_ranges should return a structure that includes both the reservation id and the range.

Alternatively, a new ip_reservation datasource could provide this functionality.  Lookups could be by IP Address + Project + Type.

Here is an example of how this could be used:

```hcl 
provider "dns" {}
provider "metal" {}

data "dns_a_record_set" "www" {
  host = "www.example.com"
}

data "metal_reserved_ip_block" "www" {
  project_id = "${var.existing_project}" // required?
  address = data.dns_a_record_set.www.addrs[0] // matches any reservation whose cidr contains this address (optional?)
  // cidr = "" // matches by cidr (optional filter?)
  // type = "public_ipv4" (optional filter?)
  // facility, metro (optional?)

  // error if multiple records match after all filters are applied?
}

resource "metal_device" "www" {
  ip_address {
    type = "public_ipv4"
     // is this an array or a string? docs need updating, they say string but the field is plural  
    reservation_ids = data.metal_reserved_ip_block.www.id  
  }
}
```

Attributes would include everything reported by a [metal_reserved_ip_block resource](https://registry.terraform.io/providers/equinix/metal/latest/docs/resources/reserved_ip_block).

t0mk:
This PR adds a datasource metal_reserved_ip_block, which can be used to look up an IP address block based on an IP address, or by the block ID. 

Like
https://registry.terraform.io/providers/linode/linode/latest/docs/data-sources/networking_ip
https://registry.terraform.io/providers/scaleway/scaleway/latest/docs/data-sources/lb_ip

We should think what we do with the metal_precreated_ip_block, which filters IP blocks by facility, metro, addr family etc. Not sure if it's useful. Is there a way to find out?

Fixes #68 

### DIFF
--- a/metal/datasource_metal_reserved_ip_block.go
+++ b/metal/datasource_metal_reserved_ip_block.go
@@ -1,0 +1,169 @@
+package metal
+
+import (
+	"fmt"
+	"net"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/packethost/packngo"
+)
+
+func dataSourceMetalReservedIPBlock() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceMetalReservedIPBlockRead,
+		Schema: map[string]*schema.Schema{
+			"block_id": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Description:   "ID of the block to look up",
+				ConflictsWith: []string{"project_id", "address"},
+				Computed:      true,
+			},
+			"project_id": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				Description:   "ID of the project where the searched block should be",
+				ConflictsWith: []string{"block_id"},
+			},
+			"ip_address": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Description:   "Find block containing this IP address in given project",
+				ConflictsWith: []string{"block_id"},
+			},
+
+			"global": {
+				Type:        schema.TypeBool,
+				Computed:    true,
+				Description: "Addresses from block are attachable in all locations",
+			},
+			"public": {
+				Type:        schema.TypeBool,
+				Computed:    true,
+				Description: "Addresses from public block are routeable from the Internet",
+			},
+			"facility": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Facility of the block. (for non-global blocks)",
+			},
+			"metro": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Metro of the block (for non-global blocks)",
+			},
+			"address_family": {
+				Type:        schema.TypeInt,
+				Computed:    true,
+				Description: "4 or 6",
+			},
+			"address": {
+				// I honestly don't know what this "address" is. Maybe next available?
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"cidr_notation": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "CIDR notation of the looked up block",
+			},
+			"cidr": {
+				Type:        schema.TypeInt,
+				Computed:    true,
+				Description: "Length of CIDR prefix of the block as integer",
+			},
+			"gateway": {
+				Type:        schema.TypeString,
+				Computed:    true,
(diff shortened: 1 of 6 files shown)


t0mk:
@displague please take a look.

t0mk:
.. I removed a read function for the old metal_precreated_ip_block datasoruce by mistake. I'm tired, I will fix it tomorrow.

metal/datasource_metal_reserved_ip_block.go
```diff
@@ -0,0 +1,169 @@
...
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/packethost/packngo"
+)
+
+func dataSourceMetalReservedIPBlock() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceMetalReservedIPBlockRead,
+		Schema: map[string]*schema.Schema{
+			"block_id": {
```
displague, line 16: Perhaps we should just use `id` here. I've done this in the past.

I don't like the name `block_id` because it doesn't match anything else (the API field name, the resource name, the API resource endpoint name).

displague: This would also be consistent with:

https://github.com/equinix/terraform-provider-metal/pull/80/files#diff-1896ab87381448e70713530c6ffa8489b7da14b7b0615646fe86bff63d4677b5R50

t0mk: @displague I think that "id" is an implicit field of any TF resource or datasource. It's always there so you don't have to define it. It's set by SetId().

The block_id is a param for resource lookup. I can rename it if you want, just suggest a different name.

example DS with similar approach:
https://registry.terraform.io/providers/scaleway/scaleway/latest/docs/data-sources/k8s_pool

t0mk:
@displague it's ready for review

metal/datasource_metal_reserved_ip_block.go
displague: The only place where we take a datasource attribute that doesn't follow "full_resource_name_id" is spot_market_request where we call it "request_id".

The only place where we use "id" as the argument is ssh_key.

The [Terraform convention](https://www.terraform.io/docs/extend/best-practices/naming.html#naming) would be to match the field name in the API so TF documentation and the API documentation can follow closely.
> Most names in a Terraform provider will be drawn from the upstream API/SDK that the provider is using

I have a hard time leaning on _convention_ when it goes against _existing practice_.

Here are some examples where I used "id" as both optional and required parameters in another provider, which matched SetId after the resource was loaded:
* https://registry.terraform.io/providers/linode/linode/latest/docs/data-sources/image
* https://registry.terraform.io/providers/linode/linode/latest/docs/data-sources/domain_record

The field name for the underlying API response for `metal_reserved_ip_block` is `id`.

We may have inherited some confusion here from the API endpoint names (`/projects/{id}/ip`) not matching
g the API documented type name (`IPReservation`).   This datasource and the resource could have been called `metal_ip` if we were starting over. 

If we want to adopt full resource names across the board, we'll have to revisit the datasources for ssh_key and spot_market_request. 

If we want to consistently adopt `id` we'll have to revisit many more data sources.

This is a tough naming call, but I think we should go with `id` for this lookup argument. It's the obvious field name if you wanted to use this resource without documentation, or your were using the EM API docs as documentation.  `block_id` is not obvious because it does not match the resource name.

It seems other providers are inconsistent in naming, fwiw:
* https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/ec2_transit_gateway#by-identifier
* https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/instance#argument-reference

displague:
Feel free to merge after the `id` field change, @t0mk 

We can then get a 2.1.0 changelog PR up and get that out before the weekend or early next week.

metal/datasource_metal_reserved_ip_block.go
t0mk: ok, I will change it to `id`

PR #81, opened 2021-04-21
add metro field to data source metal_connection (facility now optional)

displague:
Adds missing `metro` field for data source metal_connection resources.

Makes metal_connection safe for null facility (if that should be returned).


t0mk:
I tested the datasource and it populates the metro attr properly. Looks good.

PR #82, opened 2021-04-21
remove deprecation warning from metal_device.facility

### LINKED ISSUE #79: Facility is not deprecated
An unintended deprecation state made it past a previous PR review. 

Facility is not deprecated, this line should be removed:
https://github.com/equinix/terraform-provider-metal/blob/v2.0.1/metal/resource_metal_device.go#L101

Any other resources reporting that facility is deprecated should be revised as well.

displague:
This deprecation notice was accidentally added, facilities have never
been deprecated.

Fixes #79

Signed-off-by: Marques Johansson <mjohansson@equinix.com>

### DIFF
--- a/metal/resource_metal_device.go
+++ b/metal/resource_metal_device.go
@@ -98,7 +98,6 @@ func resourceMetalDevice() *schema.Resource {
 					}
 					return false
 				},
-				Deprecated:    "Use metro attribute instead",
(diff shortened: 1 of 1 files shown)


t0mk:
Looks good. Sorry, at the time I was under the impression that metros are going to substitute facilities at some point.

PR #92, opened 2021-05-03
add resource metal_virtual_circuit

t0mk:

### DIFF
--- a/docs/resources/virtual_circuit.md
+++ b/docs/resources/virtual_circuit.md
@@ -1,0 +1,54 @@
+---
+page_title: "Equinix Metal: virtual_circuit"
+subcategory: ""
+description: |-
+  Create Equinix Fabric Virtual Circuit
+---
+
+# metal_virtual_circuit
+
+Use this resource to associate VLAN with a Dedicated Port from [Equinix Fabric - software-defined interconnections](https://metal.equinix.com/developers/docs/networking/fabric/#associating-a-vlan-with-a-dedicated-port).
+
+## Example Usage
+
+Pick an existing Project and Connection, create a VLAN and use `metal_virtual_circuit` to associate it with a Primary Port of the Connection.
+
+```hcl
+locals {
+	project_id = "52000fb2-ee46-4673-93a8-de2c2bdba33c"
+	conn_id = "73f12f29-3e19-43a0-8e90-ae81580db1e0"
+}
+
+data "metal_connection" test {
+	connection_id = local.conn_id
+}
+
+resource "metal_vlan" "test" {
+	project_id = local.project_id
+	metro      = data.metal_connection.test.metro
+}
+
+resource "metal_virtual_circuit" "test" {
+	connection_id = local.conn_id
+	project_id = local.project_id
+	port_id = data.metal_connection.test.ports[0].id
+	vlan_id = metal_vlan.test.id
+	nni_vlan = 1056
+}
+```
+
+## Argument Reference
+
+* `connection_id` - (Required) UUID of Connection where the VC is scoped to
+* `project_id` - (Required) UUID of the Project where the VC is scoped to
+* `port_id` - (Required) UUID of the Connection Port where the VC is scoped to
+* `nni_vlan` - (Required) Equinix Metal network-to-network VLAN ID
+* `vlan_id` - (Required) UUID of the VLAN to associate
+* `name` - (Optional) Name of the Virtual Circuit resource
+* `facility` - (Optional) Facility where the connection will be created
+* `description` - (Optional) Description for the connection resource
+
+## Attributes Reference
+
+* `status` - Status of the virtal circuit
+* `vnid`, `nni_nvid` - VLAN parameters, see the [documentation for Equinix Fabric](https://metal.equinix.com/developers/docs/networking/fabric/)
--- a/docs/resources/virtual_circuit.md
+++ b/docs/resources/virtual_circuit.md
@@ -51,4 +51,5 @@ resource "metal_virtual_circuit" "test" {
 ## Attributes Reference
 
 * `status` - Status of the virtal circuit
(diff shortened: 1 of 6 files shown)


t0mk:
@displague I have 2 issues here:

- what we should do when user tries to create a virtual circuit for connection in status "pending", i.e. when the connections is still a request? The API will let her create a virtual circuit and assign it to a connection port, but the status of the virtual cicruit is "pending", and I don't know if it's usable or not. I could check status of the connection when creating the VC in this resource, and fail if the connection is still `pending`.

- [To create a virtual circuit, user needs to put `connectoin_id` in its creating request](https://metal.equinix.com/developers/api/connections/#connections-createConnectionPortVirtualCircuit) (correct me if I am wrong here). Right now there's no way to retrieve conenction ID from a virtual circuit. Until this is possible, the `metal_virtual_circuit` will not be importable.

displague:
Terraform resources are typical not returned until they are in a usable status. I would say holding on to the connection resource until it is ready is the Terraform approach.

How long do we have to wait for the resources to leave the pending status?

t0mk:
@displague The connection resource is in "pending" all until it's approved by EM support. I.e. all the Connection requests are connectoin resources in "pending" state. It's a bit similar as with the IP reservations.

How long we wait depends on the the approval policy for connections in the project/customer. It might also depend on the connection type and other parameters (think global/public floating IP).

t0mk:
@displague this is as good as it can get with the current state of the API. Please take a look. Also, please check the new test, it's commented out, I'm not sure whether to keep it or remove it.

displague:
I stalled on this when I noticed the following and reported it upstream:
```
2021/05/06 15:44:05 [DEBUG] DELETE https://api.equinix.com/metal/v1/connections/857e00c2-ca86-4d8e-99d9-f18f9d37f5b7
2021/05/06 15:44:05 [DEBUG] DELETE https://api.equinix.com/metal/v1/projects/53fa0f89-ad1a-4e17-9cf0-d1f09cb849c8
2021/05/06 15:44:06 [DEBUG] GET https://api.equinix.com/metal/v1/connections/857e00c2-ca86-4d8e-99d9-f18f9d37f5b7
    testing.go:630: Error destroying resource! WARNING: Dangling resources
        may exist. The full state and error is shown below.
        
        Error: Check failed: Connection still exists
        
        State: <no state>
--- FAIL: TestAccMetalConnection_Shared (3.07s)
```

After we send the `DELETE`, I assume the resource is immediately gone from view. If it is not, perhaps Terraform should poll until the resource 404s. Ideally the API should remove the resource immediately, but there may be reasons why that can not be the case.

docs/resources/virtual_circuit.md
```diff
@@ -0,0 +1,54 @@
...
+* `nni_vlan` - (Required) Equinix Metal network-to-network VLAN ID
+* `vlan_id` - (Required) UUID of the VLAN to associate
+* `name` - (Optional) Name of the Virtual Circuit resource
+* `facility` - (Optional) Facility where the connection will be created
+* `description` - (Optional) Description for the connection resource
+
+## Attributes Reference
+
+* `status` - Status of the virtal circuit
+* `vnid`, `nni_nvid` - VLAN parameters, see the [documentation for Equinix Fabric](https://metal.equinix.com/developers/docs/networking/fabric/)
```
displague, line 54: nit: these should be on two different lines

metal/resource_metal_virtual_circuit.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+	connId := d.Get("connection_id").(string)
+	portId := d.Get("port_id").(string)
+	projectId := d.Get("project_id").(string)
+
+	conn, _, err := client.Connections.Get(connId, nil)
+	if err != nil {
+		return err
+	}
+	// TODO: export consts for Connection status from packngo
```
displague, line 92: nit: TODO complete?

metal/resource_metal_virtual_circuit.go
```diff
@@ -0,0 +1,196 @@
...
+			)
+			if err != nil {
+				return 0, "", err
+			}
+			return vc, vc.Status, nil
+		},
+		Timeout:                   timeout,
+		Delay:                     10 * time.Second,
+		MinTimeout:                5 * time.Second,
+		ContinuousTargetOccurence: 5,
```
displague, line 161: what does this do and is it enough?

metal/resource_metal_virtual_circuit.go
t0mk: yes, I will remove the comment

metal/resource_metal_virtual_circuit.go
t0mk: Ah, the continuoustagetoccurence is probably not necessary here. There's no jitter in state of EM resources.

Otherwise, the statechangeconf is a struct for polling for resource state. It's the old way to do it (before contexts).

I use it here to wait for the VC state on create (just like with devices). It's also used when deleting the VC because it's necessary to detach the VLAN (which you pass in creating request of the VC).

t0mk:
> After we send the DELETE, I assume the resource is immediately gone from view. If it is not, perhaps Terraform should poll until the resource 404s. Ideally the API should remove the resource immediately, but there may be reasons why that can not be the case.

HTTP DELETE on connection has a response code too, but I don't know if it's checked as there's `ignoreResponseErrors` now.

Maybe try to remove the `ignoreResposeErrors` from the Connection Delete method. There might be some important error message in the DELETE response.

displague:
I'm suggesting that the `Delete` to Connections is working, but that we may have to wait until it is completely deleted before exiting the function.  This is similar to how we need to wait until a Device is fully provisioned before handing control over to Terraform.

The reason why this seems to be necessary in Delete, is that the Delete is not immediately enacted. We don't want a seemingly unrelated action, like deleting a Project, can fail if the cause is that one of our resource methods exited too early.

displague:
We can address the Connection deletion concerns in a follow-up PR, if needed/possible.

PR #96, opened 2021-05-06
metal_port datasource

### LINKED ISSUE #85: Create a metal_port datasource
Introducing `metal_port` as a datasource, ahead of any refactoring to offer a `metal_port` resource (https://github.com/equinix/terraform-provider-metal/issues/12#issuecomment-820444150) will give users the ability to work with ports that may be in use by appliances, connections, or devices.  The attributes of this datasource will give us a head start on a `metal_port` resource.

This data source could offer `device_id` search fields in addition to `id` (port_id).  Port name may also be a helpful search field. 

For users looking to fetch the `metal_device_network_type`  of a device, this would be possible by fetch the network type on each port and performing the same logic used by `metal_device_network_type` (this logic is not present in the EM API, this TF provider formulates it, like the EM Portal.  This single device "network_type" does not fit the broad reality of port configurations across all EM device models.

Users can get to the `device_id` through the existing `metal_vlan` data source.  Likewise, `ports` can be listed on the `metal_device` datasource.  However, there is no TF way (that I know of now) to get the ports related to a VLAN (or vice-versa) today.

This resource should follow `/ports/{id}`.  There may be other port attributes exposed in devices, connections, VLAN, or other resources, but my first inclination is to avoid including those in `metal_ports`. If we've omitted useful fields in those other resources, that are within a `ports` attribute, we should fill in the missing attributes in those resources.

t0mk:
fixes #85 

This PR adds a ds for metal_port, lookup of ports by ID or device ID and port name.

### DIFF
--- a/docs/data-sources/port.md
+++ b/docs/data-sources/port.md
@@ -1,0 +1,45 @@
+---
+page_title: "Equinix Metal: precreated_port"
+subcategory: ""
+description: |-
+  Fetch device ports
+---
+
+# metal_port
+
+Use this data source to read ports of existing devices. You can read port by either its UUID, or by a device UUID and port name.
+
+## Example Usage
+
+Create a device and read it's eth0 port to the datasource.
+
+```hcl
+locals {
+  project_id = "<UUID_of_your_project>"
+}
+
+resource "metal_device" "test" {
+  hostname         = "tfacc-test-device-port"
+  plan             = "c3.medium.x86"
+  facilities       = ["sv15"]
+  operating_system = "ubuntu_20_04"
+  billing_cycle    = "hourly"
+  project_id       = local.project_id
+}
+
+data "metal_port" "test" {
+    device_id        = metal_device.test.id
+    port_name        = "eth0"
+}
+```
+
+## Argument Reference
+
+* `id` - (Required) ID of the port to read, conflicts with device_id.
+* `device_id` - (Required) 4 or 6, depending on which block you are looking for.
+* `port_name` - (Required) Whether to look for public or private block.
+
+## Attributes Reference
+
+* `bond_port` - Whether port is a bond port or physical port
+* `network_type` - One of layer2-bonded, layer2-individual, layer3, hybrid, hybrid-bonded
--- a/docs/data-sources/port.md
+++ b/docs/data-sources/port.md
@@ -28,18 +28,26 @@ resource "metal_device" "test" {
 }
 
 data "metal_port" "test" {
-    device_id        = metal_device.test.id
-    port_name        = "eth0"
+    device_id = metal_device.test.id
+    name      = "eth0"
 }
 ```
 
 ## Argument Reference
 
 * `id` - (Required) ID of the port to read, conflicts with device_id.
-* `device_id` - (Required) 4 or 6, depending on which block you are looking for.
-* `port_name` - (Required) Whether to look for public or private block.
+* `device_id` - (Required) 
+* `name` - (Required) Whether to look for public or private block.
 
 ## Attributes Reference
 
-* `bond_port` - Whether port is a bond port or physical port
 * `network_type` - One of layer2-bonded, layer2-individual, layer3, hybrid, hybrid-bonded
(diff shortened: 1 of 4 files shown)


docs/data-sources/port.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Argument Reference
+
+* `id` - (Required) ID of the port to read, conflicts with device_id.
+* `device_id` - (Required) 4 or 6, depending on which block you are looking for.
+* `port_name` - (Required) Whether to look for public or private block.
+
+## Attributes Reference
+
+* `bond_port` - Whether port is a bond port or physical port
```
displague, line 44: `bond_port` here is a TF boolean based on the `type` field matching `NeworkBondPort`.  I think we should just expose `type` and let the user apply this logic if they need it.  The `type` field could present new options in the future.

There are a few more attributes that we can expose here: https://github.com/packethost/packngo/blob/master/ports.go#L45-L75
* `type`
* `data` { mac, bonded }
* `disbond_operation_supported`
* `native_virtual_network`
* `virtual_networks`
* `bond` { id, name }

docs/data-sources/port.md
```diff
@@ -0,0 +1,45 @@
...
+    device_id        = metal_device.test.id
+    port_name        = "eth0"
+}
+```
+
+## Argument Reference
+
+* `id` - (Required) ID of the port to read, conflicts with device_id.
+* `device_id` - (Required) 4 or 6, depending on which block you are looking for.
+* `port_name` - (Required) Whether to look for public or private block.
```
displague, line 40: Perhaps we should use the name `name`, to match the API.

docs/data-sources/port.md
```diff
@@ -0,0 +1,45 @@
...
+data "metal_port" "test" {
+    device_id        = metal_device.test.id
+    port_name        = "eth0"
+}
+```
+
+## Argument Reference
+
+* `id` - (Required) ID of the port to read, conflicts with device_id.
+* `device_id` - (Required) 4 or 6, depending on which block you are looking for.
```
displague, line 39: I considered suggesting that we call this `hardware` to match the API, but I'm not sure how we would be able to determine if the user means "hardware is a reservation" or "hardware is a device instance" or "hardware is something else".

It makes sense to use "device_id" + "port.name" for lookups, and if we need more lookups in the future we'll have to add more lookup fields.

Should we set `device_id` based on the `hardware.href` returned (assuming there href basename is the id)?  Perhaps we need to query `/ports/{id}?include=hardware` to gather the needed information.

docs/data-sources/port.md
displague: Other concerns aside, I'd be glad to merge this sooner and add those fields in a follow-up.

metal/datasource_metal_port_test.go
```diff
@@ -0,0 +1,94 @@
...
+	return fmt.Sprintf(`
+
+resource "metal_project" "test" {
+    name = "tfacc-port-%s"
+}
+
+resource "metal_device" "test" {
+  hostname         = "tfacc-test-device-port"
+  plan             = "c3.medium.x86"
+  facilities       = ["sv15"]
```
displague, line 40: Unless we are testing `facility` workflows, we should test with metros for a better success rate in tests (more availability).

docs/data-sources/port.md
t0mk: sorry about the doc string for the `device_id`.

I'm a bit confused here. The hardware ID is different than the device ID, and the href link in hardware is not GETtable. I get 404 getting e..g /hardware/8eec90d7-62c4-47c2-b541-b0a0a3f1c428.

I can see the hardware resource when I GET "/ports/9ea27e03-8c5f-4a2a-ac6e-466cb1076c36?include=hardware". It has the device in `instance` attribute. However, we don't have a struct for the "hardware" resource here or in packngo.

docs/data-sources/port.md
t0mk: ok

metal/datasource_metal_port_test.go
t0mk: :+1:

PR #100, opened 2021-05-10
Add datasource metal_hardware_reservation and docs

### LINKED ISSUE #86: Create a metal_hardware_reservation data source
There is currently no way to access a matching hardware reservation record in the TF provider.  What criteria should be permitted aside from hardware_reservation_id?

It should also be possible to query metal_devices by hardware_reservation_id.

t0mk:
fixes #86 

This PR adds datasource metal_hardware_reservation. 

@displague please let me know if we should allow to look up the reservation by other params than the ID.

Also, please fix some docstrings in the code, I wasn't sure about the purpose of some parameters. Grep for "@displague"

### DIFF
--- a/docs/data-sources/hardware_reservation.md
+++ b/docs/data-sources/hardware_reservation.md
@@ -1,0 +1,36 @@
+---
+page_title: "Equinix Metal: metal_hardware_reservation"
+subcategory: ""
+description: |-
+  Retrieve Equinix Metal Hardware Reservation
+---
+
+# metal_hardware_reservation
+
+Use this data source to retrieve a [hardware reservation resource from Equinix Metal](https://metal.equinix.com/developers/docs/deploy/reserved/).
+
+## Example Usage
+
+```hcl
+data "hardware_reservation" "example" {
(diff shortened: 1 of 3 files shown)


docs/data-sources/hardware_reservation.md
```diff
@@ -0,0 +1,36 @@
...
+
+```hcl
+data "hardware_reservation" "example" {
+  id     = "4347e805-eb46-4699-9eb9-5c116e6a0172"
+}
+```
+
+## Argument Reference
+
+* `id` - (Required) ID of the hardware reservation
```
displague, line 22: `device_id` may be the most helpful lookup after `id`. We can add that in a separate PR (we'll need an issue to track).

t0mk:
@displague I fixed the doc strings as suggested. I also did the lookup by device_id. I had to fix the type of the HardwareReservation attribute of packngo.Device:
https://github.com/packethost/packngo/pull/282

The code now doens't compile. I will need to bump packngo and then rebase.

t0mk:
@displague I fixed the doc strings as suggested, and I also did a lookup by device_id. It's ready for review again.

displague (approved):

PR #108, opened 2021-05-20
prepare changelog for 3.0.0

displague:
updates CHANGELOG for 3.0.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,45 @@
+## 3.0.0 (May 20, 2021)
+
+BREAKING CHANGES:
+- `metal_spot_market_request` field `locked` is now boolean (not string) ([#78](https://github.com/equinix/terraform-provider-metal/pull/78))
+
+DEPRECATIONS:
+- **Deprecated Resource:** `metal_volume` ([#99](https://github.com/equinix/terraform-provider-metal/pull/99))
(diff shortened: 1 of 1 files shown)


displague:
On second thought - let's release this as 2.1.0. The breaking change is minor and easy to address in configurations.

Other providers introduce similar breaking changes with only minor version bumps:
https://github.com/hashicorp/terraform-provider-google/blob/master/CHANGELOG.md

This is consistent with TF best practices:
https://www.terraform.io/docs/extend/best-practices/versioning.html#versioning-specification

Specifically, ["Changing attribute type where the new type is functionally compatible (e.g. TypeInt to TypeFloat)"](https://www.terraform.io/docs/extend/best-practices/versioning.html#example-minor-number-increments).

`"true"` / `"false"` was previously required. These values will translate safely to boolean.

---
The following demonstrates:

```hcl
# main.tf
variable "t" {
 type = string
}
module "x" {
  source = "./module/"
  t = var.t
}
output "t" {
 value = module.x.t
}
```

```hcl
# module/main.tf
variable "t" {
 type = bool
}
output "t" {
 value = var.t
}
```

```hcl
# terraform.tfvars
t = "true"
```

```sh
$ terraform init
...
$ terraform apply
...
Outputs:
t = true
```

PR #111, opened 2021-05-21
Fix typos in the new guides

t0mk:
I went trough the rendered guides in TF doc site:
https://registry.terraform.io/providers/equinix/metal/latest/docs/guides/migration_guide_packet
.. and fixed some typos.

### DIFF
--- a/docs/guides/migration_guide_facilities_to_metros_devices.md
+++ b/docs/guides/migration_guide_facilities_to_metros_devices.md
@@ -8,7 +8,7 @@ description: |-
 
 In April 2021, Equinix Metal rolled out a new location concept - [metros](https://feedback.equinixmetal.com/changelog/new-metros-feature-live). A metro is an Equinix-wide concept for data centers which are grouped together geographically. Data centers within a metro share capacity and networking features. You can read more about metros at https://metal.equinix.com/developers/docs/locations/metros/.
 
-Until the metros introduction, resource deployment location used to be controlled by "facilty" - single data center with location code like "sv15" or "ny5". Metros group the facilities, and e.g. metro "sv" contains facility "sv15". If you specify a metro when creating a resource, it will be deployed to one of the facilities in the metro group. You can then find the deployed facility from the a read-only attribute of the resource.
+Until the metros introduction, resource deployment location used to be controlled by "facilty" - single data center with location code like "sv15" or "ny5". Metros group the facilities, so e.g. metro "sv" contains facility "sv15". If you specify a metro when creating a resource, it will be deployed to one of the facilities in the metro group. You can then find the deployed facility from a read-only attribute of the resource.
 
 
 ## Changing your Terraform templates to use metros instead of facilities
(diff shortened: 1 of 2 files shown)


docs/guides/migration_guide_facilities_to_metros_devices.md
```diff
@@ -8,7 +8,7 @@ description: |-
...
 
-Until the metros introduction, resource deployment location used to be controlled by "facilty" - single data center with location code like "sv15" or "ny5". Metros group the facilities, and e.g. metro "sv" contains facility "sv15". If you specify a metro when creating a resource, it will be deployed to one of the facilities in the metro group. You can then find the deployed facility from the a read-only attribute of the resource.
+Until the metros introduction, resource deployment location used to be controlled by "facilty" - single data center with location code like "sv15" or "ny5". Metros group the facilities, so e.g. metro "sv" contains facility "sv15". If you specify a metro when creating a resource, it will be deployed to one of the facilities in the metro group. You can then find the deployed facility from a read-only attribute of the resource.
```
displague, line 11:
```suggestion
Before the introduction of metros, resources were deployed to a single `facility` location.  When provisioning `metal_device` resources, the facility could be chosen by Equinix Metal with a user-supplied list of `facilities`, or a wildcard `any` facility.  The individual facility locations use a code like "sv15" or "ny5". Metros group facilities. For example, metro "sv" contains the "sv15" facility among others. If you specify a metro when creating a resource, it will be deployed to one of the facilities in the metro group. You can then find the deployed facility using a read only attribute of the resource (e.g. `deployed_facility` for `metal_device` resources).
```

docs/guides/migration_guide_facilities_to_metros_devices.md
```diff
@@ -48,9 +48,9 @@ resource "metal_device" "node" {
 }
 ```
 
-To test that the change didn't taint the state, and that the device will not be re-created, you can check if `terraform plan` reports no differences. The terraform state should be up to date as long as the facility in which the device was deployed was contained within the metro.
+To test that the change didn't taint the state, and that the device will not be re-created, you can check if `terraform plan` reports any differences. The terraform state should be up to date as long as the facility in which the device was deployed was contained within the metro.
 
-If the plan diff is not empty, you might have set a metro not containing the facility to which the device was deployed. This might happen if you've used more facilities in the `facilities` list, or you have used "any" facility.
+If the plan diff is not empty, you might have used a metro not containing the facility to which the device was deployed. This might happen if you've used more facilities in the `facilities` list, or you have used "any" facility.
```
displague, line 53:
```suggestion
If the plan diff is not empty, you might have used a metro not containing the facility to which the device was deployed. This might happen if you've used more facilities in the `facilities` list, or you have used the special "any" facility.
```

docs/guides/migration_guide_packet.md
```diff
@@ -97,11 +97,11 @@ If the plan is not empty, it means that some resources can not be simply read fo
 
 ## Migrating one resource at a time
 
-We can use terraform state and import to achieve transition without destroying existing resources.
+We can use `$ terraform state` and `$ terraform import` to achieve transition without destroying existing resources.
```
displague, line 100: no need for `$ ` here.
```suggestion
We can use `terraform state` and `terraform import` to achieve transition without destroying existing resources.
```

displague:
Unrelated note, but I noticed that if I have modules in my project, using the `packet` provider, I also had to add a `required_providers` block in that module:

```
terraform {
  required_providers {
    metal = {
      source  = "equinix/metal"
      version = "~> 2.1"
    }
}
```

If I didn't I got errors about `hashicorp/metal` not being found.

t0mk:
> Unrelated note, but I noticed that if I have modules in my project, using the `packet` provider, I also had to add a `required_providers` block in that module:
> 
> ```
> terraform {
>   required_providers {
>     metal = {
>       source  = "equinix/metal"
>       version = "~> 2.1"
>     }
> }
> ```
> 
> If I didn't I got errors about `hashicorp/metal` not being found.

The `required_providers` block is in the templates in the migration guide from the old Packet provider. It's not in the guide for migrating devices from fac to metro, because the templates are not complete, it just shows the required change in the device resoruces.

But

displague (approved):

PR #113, opened 2021-05-24
Terraform Plugin SDK v2 Upgrade

### LINKED ISSUE #77: Update to Terraform SDK 2
> New major version is out and all providers should be updated. See https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html for more details.

_Originally https://github.com/packethost/terraform-provider-packet/issues/286 from @invidian_

t0mk:
This PR updates the plugin sdk to v2. I had to comment out usage of mutexkv and terraform sdk httpclient. I will need to find ways to redo it properly before we merge.

fixes #77 

reference:
https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html


t0mk:
I fixed the mutexkv and the user agent string.

t0mk:
@displague please take a look. Sorry that the diff is a bit messy.

metal/datasource_metal_device.go
```diff
@@ -93,11 +93,7 @@ func dataSourceMetalDevice() *schema.Resource {
 				Computed: true,
 			},
 			"storage": {
-				Type: schema.TypeString,
-				StateFunc: func(v interface{}) string {
-					s, _ := structure.NormalizeJsonString(v)
```
displague, line 98: We should make an issue to convert this from JSON to a structured value. I think we should make this match the CPR storage value submitted to the API, if possible.

metal/mutexkv.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"log"
+	"sync"
+)
+
+// MutexKV is a simple key/value store for arbitrary mutexes. It can be used to
+// serialize changes across arbitrary collaborators that share knowledge of the
+// keys they must serialize on.
+//
+// The initial use case is to let aws_security_group_rule resources serialize
```
displague, line 12: Perhaps we could change this to note why we are using it in this provider:
```
               // Equinix Metal doesn't allow multiple VLANs to be assigned
                // to the same port at the same time
                lockId := "vlan-attachment-" + port.ID
                metalMutexKV.Lock(lockId)
                defer metalMutexKV.Unlock(lockId)
```

I wonder if we really need this. A mutex lock could be used in other places, I think, but perhaps a backoff-retry loop  would be sufficient and save us the complexity of locks.

Is this even more important in newer TF because of provider aliases?

metal/resource_metal_device.go
```diff
@@ -280,7 +280,7 @@ func resourceMetalDevice() *schema.Resource {
 					s, _ := structure.NormalizeJsonString(v)
```
displague, line 280: Curious why we removed `NormalizeJsonString` from the data source, but left it here. I don't know why this was happening or would need to change. Just curious.

metal/mutexkv.go
displague: looks like the latest text failed related to these changes:
`metal/provider.go:9:20: undefined: mutexkv`

metal/datasource_metal_device.go
t0mk: https://github.com/equinix/terraform-provider-metal/issues/115

metal/mutexkv.go
t0mk: > A mutex lock could be used in other places, I think, but perhaps a backoff-retry loop would be sufficient and save us the complexity of locks.

I really think that a mutex lock and unlock is much more simple than a retry loop. It's just 3 lines, no error checking, no API calls.

> Is locking even more important in newer TF because of provider aliases?

I don't understand how provider aliases are related. The locking here is to prevent TF parallelism to ruin an apply run - TF runs several goroutines and parallelizes everyhting that is not dependent. I think a HC eng did this after acceptance test were failing because of parallel port-vlan assignment. I am not sure if the mutex map is the best solution, but we didn't have an issue with the parallel assignment since.

I had to drop the mutex map code because it was removed in the SDK v2.
https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html#removal-of-helper-mutexkv-package

metal/resource_metal_device.go
t0mk: There was an internal check in the provider that failed on the StateFunc because it doesn't have sense in computed attribute:
https://github.com/bflad/tfproviderlint/tree/main/passes/S033

I left it in the Read function because it's better to have the JSON normalized.

displague:
All of the debug messages are including this bit about the User-Agent. I'm curious if that is expected and also curious about why it is not showing our provider name/version in the string.

The 504 is also interesting, this would have been a good error for the client to retry.

```
    resource_metal_port_vlan_attachment_test.go:229: Step 2/2 error: Error running apply: exit status 1
        2021/05/24 16:02:42 [DEBUG] Using modified User-Agent: Terraform/0.12.29 HashiCorp-terraform-exec/0.13.0
        
        Error: POST https://api.equinix.com/metal/v1/ports/6e0b7338-c748-4cea-ab23-41378648a7fd/disbond: 504  Unexpected Content-Type text/html with status 504 Gateway Time-out
```

metal/mutexkv.go
displague: I assumed the mutex would be shared between all aliases of the provider, but now I'm not sure about that. With aliases perhaps each instance is executing a separate copy of the plugin binary.

A retry loop would prevent cases where multiple clients (not just terraform, or multiple copies of terraform) are running against the same account. Maybe something to consider in the future, maybe not.

Agreed - Updating the mutex for sdkv2 sounds good.

t0mk:
> All of the debug messages are including this bit about the User-Agent.

That was another fix of missing package:
https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html#removal-of-httpclient-package

I chose to replicate the v1 user agent
https://github.com/equinix/terraform-provider-metal/pull/113/files#diff-e782a70954f139f5a8c0f513107ad60eca5cd104f862ea932921c32773636b79R87
.. in the same way like in the Linode provider to keep the UserAgent format similar in case it's recorded in the API.

t0mk:
>         Error: POST https://api.equinix.com/metal/v1/ports/6e0b7338-c748-4cea-ab23-41378648a7fd/disbond: 504  Unexpected Content-Type text/html with status 504 Gateway Time-out

I would be careful with retries on HTTP errors, IMO it can do more harm than good. For user experience it's better to fail early than retry on timeouts. Anyway, the content-type of that error shouldn't be text/html.

displague:
The user-agent does say anything about Equinix Metal or packngo, is that expected? I believe we previously had those details in the user-agent.

> Providers are encouraged to construct User-Agent strings to suit their needs, as they know best how to represent their provider to their APIs.

t0mk:
@displague can you verify? I've just run project acc test

```
tomk@xps ~/tpm/metal ±tf-sdk-v2 » TF_LOG=DEBUG TF_ACC=1 go test -v -timeout=20m -run=TestAccMetalProject_Basic
=== RUN   TestAccMetalProject_Basic


2021/05/31 14:53:48 [DEBUG] POST https://api.equinix.com/metal/v1/projects
2021/05/31 14:53:48 [DEBUG] Equinix Metal API Request Details:
---[ REQUEST ]---------------------------------------
POST /metal/v1/projects HTTP/1.1
Host: api.equinix.com
User-Agent: HashiCorp Terraform/0.11+compatible (+https://www.terraform.io) Terraform Plugin SDK/2.4.3 terraform-provider-metal/dev packngo/0.14.1
[...]
```

.. but I'm a bit alarmed that I can see the Auth token header in the debug log of the HTTP requeset. I thought we've fixed that somehow at some point in the past. I will try to fix it.

displague:
The user-agent looks good.

>  but I'm a bit alarmed that I can see the Auth token header in the debug log of the HTTP requeset. I thought we've fixed that somehow at some point in the past. I will try to fix it.

Packngo is redacting this (run with `PACKNGO_DEBUG=1` and you will see).  What is not redacted is a copy of the request that Terraform is emitting. https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/logging/transport.go#L19

I couldn't find examples of providers changing this behavior.

PR #126, opened 2021-06-13
fix metro statefuncs in datasources, add metro statefunc to ip resources

### LINKED ISSUE #125: Internal Validation errors after sdk v2 merge
```
        Error: InternalValidate
        
        Internal validation of the provider failed! This is always a bug
        with the provider itself, and not a user issue. Please report
        this bug:
        
        5 errors occurred:
        	* data source metal_facility: metro: StateFunc is extraneous, value should
        just be changed before setting on computed-only field
        	* data source metal_device: metro: StateFunc is extraneous, value should just
        be changed before setting on computed-only field
        	* data source metal_reserved_ip_block: metro: StateFunc is extraneous, value
        should just be changed before setting on computed-only field
        	* data source metal_connection: metro: StateFunc is extraneous, value should
        just be changed before setting on computed-only field
        	* data source metal_spot_market_request: metro: StateFunc is extraneous,
        value should just be changed before setting on computed-only field

```

t0mk:
Fixes internal validations errors caused by metro lowercasing and sdk-v2 upgrade, fixes #125 

### DIFF
--- a/metal/datasource_metal_connection.go
+++ b/metal/datasource_metal_connection.go
@@ -1,6 +1,8 @@
 package metal
 
 import (
+	"strings"
+
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/packethost/packngo"
 )
@@ -161,7 +163,7 @@ func dataSourceMetalConnectionRead(d *schema.ResourceData, meta interface{}) err
 		d.Set("facility", conn.Facility.Code)
 	}
 	if conn.Metro != nil {
-		d.Set("metro", toLower(conn.Metro.Code))
+		d.Set("metro", strings.ToLower(conn.Metro.Code))
 	}
 	d.Set("token", conn.Token)
 	d.Set("type", conn.Type)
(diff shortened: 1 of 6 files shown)


t0mk:
@displague please check this, it would be good to have it in before `metal_port` merge.

metal/datasource_metal_connection.go
```diff
@@ -162,7 +161,7 @@ func dataSourceMetalConnectionRead(d *schema.ResourceData, meta interface{}) err
 		d.Set("facility", conn.Facility.Code)
 	}
 	if conn.Metro != nil {
-		d.Set("metro", conn.Metro.Code)
+		d.Set("metro", toLower(conn.Metro.Code))
```
displague, line 164: perhaps we should just use `strings.ToLower` and avoid the `interface{}` wrapping.

metal/datasource_metal_facility.go
```diff
@@ -59,7 +58,7 @@ func dataSourceMetalFacilityRead(d *schema.ResourceData, meta interface{}) error
 				"features": f.Features,
 				"metro": func(d *schema.ResourceData, k string) error {
 					if f.Metro != nil {
-						return d.Set(k, f.Metro.Code)
+						return d.Set(k, toLower(f.Metro.Code))
```
displague, line 61: same as https://github.com/equinix/terraform-provider-metal/pull/126/files#r650953584

I think `toLower` makes sense for the statefuncs, but it adds indirection in the `d.Set` arguments.

t0mk: OK, I changed toLower to strings.ToLower in relevant places in the code

Issue #13, opened 2020-12-17
Generate docs with go generator


displague:
It is now possible to generate provider docs using the https://www.terraform.io/docs/registry/providers/docs.html#generating-documentation generator and fields embedded in the provider.

Can we generate all of our documentation identically in this way?

The generator is noted to have some weakness handling deep structures.

t0mk:
I have just tried `tfplugindocs` in this provider. It nicely generated all the markdown files to the docs/ dir.

However, to take full advantage of `tfplugindocs` (to have docs build with the provider) we'd need to
- copy description from our written docs to Description attr of each field of the resource schema
- update terraform-plugin-sdk to (at least) newest 1x version, so that we can add Description field of schema.Resource for every resource (and I guess datasource too)
- create examples/{data-sources,provider,resources} dirs with contents described in https://github.com/hashicorp/terraform-plugin-docs#conventional-paths
- reformat all the examples from the current docs into files like e.g. examples/resources/metal_device/resource.tf
- add tfplugindocs as "tools model" dependency, i.e. in `tools/tools.go`:
  ```go
  // +build tools
  
  package tools

  //go:generate go install github.com/hashicorp/terraform-plugin-docs/cmd/tfplugindocs

  import (
	// docs generator
	_ "github.com/hashicorp/terraform-plugin-docs/cmd/tfplugindocs"
  )
  ```
- add `//go:generate go run github.com/hashicorp/terraform-plugin-docs/cmd/tfplugindocs` to `<repo_dir>/main.go`

PR #130, opened 2021-06-17
Description attributes to resource structs

t0mk:
This PR copies attribute descriptions from docs to golang code, so that we can benefit from automatic doc generation.

Towards #13


t0mk:
@displague this is the first step to generating docs

I would like this to be merged rather sooner than later, so that there's no drift.

When this is merged, we can start moving docs content to [conventional paths](https://github.com/hashicorp/terraform-plugin-docs#conventional-paths).

metal/datasource_metal_device_bgp_neighbors.go
```diff
@@ -9,48 +9,58 @@ func bgpNeighborSchema() *schema.Resource {
...
+				Description: "Array of IP addresses of this neighbor's peers",
+				Optional:    true,
+				Elem:        &schema.Schema{Type: schema.TypeString},
 			},
 			"routes_in": {
-				Type:     schema.TypeList,
-				Computed: true,
-				Elem:     bgpRouteSchema(),
+				Type:        schema.TypeList,
+				Description: "Array of incoming routes. Each route has attributes:",
```
displague, line 55: was this a multiline description that got cut in the move?

metal/datasource_metal_device_bgp_neighbors.go
```diff
@@ -60,12 +70,14 @@ func bgpRouteSchema() *schema.Resource {
...
-				Computed: true,
+				Type:        schema.TypeString,
+				Description: "CIDR expression of route (IP/mask)",
+				Computed:    true,
 			},
 			"exact": {
-				Type:     schema.TypeBool,
-				Computed: true,
+				Type:        schema.TypeBool,
+				Description: "(bool) Whether the route is exact",
```
displague, line 79: "(bool) ", we won't need that.

metal/resource_metal_device_network_type.go
```diff
@@ -20,12 +20,14 @@ func resourceMetalDeviceNetworkType() *schema.Resource {
...
-				Required: true,
-				ForceNew: true,
+				Type:        schema.TypeString,
+				Description: "The ID of the device on which the network type should be set",
+				Required:    true,
+				ForceNew:    true,
 			},
 			"type": {
 				Type:         schema.TypeString,
+				Description:  "Network type to set. Must be one of layer3, hybrid, layer2-individual and layer2-bonded",
```
displague, line 30: Perhaps in the future, we can define lists like this as a const and then use a function to strings.Join (with "Must be one of ... or ...") for Description while reusing the const for ValidateFunc.

displague (approved):

displague:
I merged #126 first and that introduced a conflict here.

After resolving that (and double-checking that any multiline descriptions are retained, like the one I left a comment about), this looks good to merge.

t0mk:
@displague thanks for taking a look on this, it's a long diff and not very interesting.. I will resolve the conflict, fix your comment, check on more time and then merge.

metal/datasource_metal_device_bgp_neighbors.go
t0mk: yes

metal/resource_metal_device_network_type.go
t0mk: I can do it here and now.

PR #133, opened 2021-06-19
reserved_ip_block: add support for setting tags

### LINKED ISSUE #128: Implement tags for metal_reserved_ip_block
Originally https://github.com/packethost/terraform-provider-packet/issues/298

Implemented in https://github.com/packethost/terraform-provider-packet/pulls/300

flokli:
The API allows specifying tags for reserved IP blocks, but this wasn't
exposed in the terraform resource.

Adding this makes it possible to create a IP reservation from terraform,
to be consumed by
https://github.com/equinix/cloud-provider-equinix-metal, for its control
plane elastic IP feature.

Closes #128.

### DIFF
--- a/docs/resources/reserved_ip_block.md
+++ b/docs/resources/reserved_ip_block.md
@@ -93,6 +93,7 @@ The following arguments are supported:
 * `facility` - (Optional) Facility where to allocate the public IP address block, makes sense only for type==public_ipv4, must be empty for type==global_ipv4, conflicts with `metro`
 * `metro` - (Optional) Metro where to allocate the public IP address block, makes sense only for type==public_ipv4, must be empty for type==global_ipv4, conflicts with `facility`
 * `description` - (Optional) Arbitrary description
+* `tags` - (Optional) String list of tags
 
 ## Attributes Reference
 
@@ -110,6 +111,7 @@ The following attributes are exported:
 * `address_family` - Address family as integer (4 or 6)
 * `public` - boolean flag whether addresses from a block are public
 * `global` - boolean flag whether addresses from a block are global (i.e. can be assigned in any facility)
+* `tags` - string list of tags
 
 Idempotent reference to a first "/32" address from a reserved block might look like this:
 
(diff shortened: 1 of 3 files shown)


displague:
Thanks, @flokli!  Have a look at https://github.com/packethost/terraform-provider-packet/pull/300/files to see how this was previously implemented by @t0mk. That PR didn't make it into this repo as we were moving things from Packet to Equinix Metal, but it looks to be largely complete and compatible.

Notably, in that PR we went with `TypeSet` over `TypeList`, because tag order does not impact the resource: https://www.terraform.io/docs/extend/schemas/schema-types.html#typeset. 

You'll also find a test that can be lifted and shifted.

displague:
I see that you took the `TypeList` approach already taken by `metal_device`, which seems quite reasonable.
https://github.com/equinix/terraform-provider-metal/blob/main/metal/resource_metal_device.go#L440-L443

The only other resource that includes tags today is the SpotMarket resource which largely follows the device implementation.

I've opened #134 to reach API parity for other resources that support `tag` fields.

displague:
We can see that Hashicorp's engineers have chosen TypeSet in providers that they maintain through code generation:
https://github.com/hashicorp/terraform-provider-google-beta/blob/19dd6481f0cd0a76dbb19e33f1387aa45a321482/google-beta/resource_compute_instance.go#L672-L678

The `Set` property value should also be taken into consideration:
https://pkg.go.dev/github.com/hashicorp/terraform/helper/schema#HashString

flokli:
So should I actually keep this PR as-is, and only add some tests?

Or should I lift-and-shift the approach from the old PR you previously linked to?
That's pushed to my `reserved-ip-block-tags-2` branch.

displague:
Let's take the `TypeSet` approach in this PR.  I would just adapt this PR as needed.

displague:
https://github.com/equinix/terraform-provider-metal/compare/main...flokli:reserved-ip-block-tags-2?expand=1 looks good

flokli:
Alright, I'll push it over to here in a second.

flokli:
It's pushed. I didn't run the acceptance tests, just ran a `terraform plan` in the same workspace as the previous iteration of this PR, and it reported no changes there, so I think it's fine.

t0mk:
@flokli looks good. I just fixed the test a bit. The Public test succeeds now.

flokli:
Thanks for merging this so quickly!

PR #137, opened 2021-06-22
bump packngo to 0.16.0

t0mk:
This PR bump packgno to 0.16.0 in the provider


t0mk:
@displague do we have to do sth about the version in packngo? I run acceptance test and it reports User Agent as

HashiCorp Terraform/0.11+compatible (+https://www.terraform.io) Terraform Plugin SDK/2.4.3 terraform-provider-metal/dev packngo/(devel)

t0mk:
Also, I guess we're not 0.11 compatible anymore :D

t0mk:
https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html#removal-of-httpclient-package

displague:
> @displague do we have to do sth about the version in packngo?

https://github.com/packethost/packngo/pull/281 should take care of this. I tested it with packet-cli locally (although the package hasn't been updated for packngo 0.15+).

Is your stage dirty? I see some issues that suggest this could lead to `(devel)`. 

Perhaps the vendors/ directory causes a different behavior.

t0mk:
@displague I think the bad UserAgent might be because I got the HTTP dump from acceptance tests .. let me build manually and check again.

t0mk:
hmm, when I did the build, it shows 
```
HashiCorp Terraform/0.11+compatible (+https://www.terraform.io) Terraform Plugin SDK/2.4.3 terraform-provider-metal/dev packngo/v0.16.0
```

.. so packngo version is good, but TF version and tf provider version are not. I have TF 0.14 installed.

But maybe this is not important right now, we can create and issue to fix it. Let's merge this so we can move on with the connection flag.

PR #152, opened 2021-07-14
feat: reinstall block for devices

rawkode:
If the `reinstall { enabled = true }` HCL exists for a device and the operating system, userdata, or custom data are updated; then we'll issue a reinstall action rather than a destroy / create.

Notes for @t0mk and @displague:

I think I've taken this as far as I can and I would appreciate some assistance.

## Problem 1

The 5th test-case in `_Update` seems to be wrong: `Config: testAccCheckMetalDeviceConfig_userdata(rInt+4, rs),`

This is updating the user-data and expecting the same device, which isn't the expected or actual behaviour.

Edit: Looks like a rebase removed this test-case, so I'm assuming someone already noticed it was incorrect? :)

## Problem 2

My code appears to work. The challenge is the asynchronous reinstall state; the provider tests seem to run `terraform plan` after and it can see the device is in the `reinstalling` state with the `deprovision` OS; so it thinks that there is stuff to be done. I don't know how to get aroundm this?

Edit: Resolved with @displague. Added `waitForActiveDevice()`

### DIFF
--- a/metal/resource_metal_device.go
+++ b/metal/resource_metal_device.go
@@ -379,33 +379,37 @@ func resourceMetalDevice() *schema.Resource {
 			},
 		},
 		CustomizeDiff: customdiff.Sequence(
-			customdiff.ForceNewIf("reinstall", func(_ context.Context, d *schema.ResourceDiff, meta interface{}) bool {
-				reinstall, ok := d.GetOk("reinstall")
-
-				// Prior behaviour was to always destroy and create,
-				// so in the event we can't get the reinstall config; let's
-				// continue to do so.
-				if !ok {
-					return true
-				}
-
-				// We didn't get a reinstall configuration
-				reinstall_list, ok := reinstall.([]interface{})
-				if !ok {
-					return true
-				}
-
-				reinstall_config, ok := reinstall_list[0].(map[string]interface{})
-
-				// We didn't get a reinstall configuration
-				if !ok {
-					return true
-				}
-
-				return !reinstall_config["enabled"].(bool)
-			}),
+			customdiff.ForceNewIf("custom_data", shouldReinstall),
+			customdiff.ForceNewIf("operating_system", shouldReinstall),
+			customdiff.ForceNewIf("user_data", shouldReinstall),
 		),
 	}
+}
+
+func shouldReinstall(_ context.Context, d *schema.ResourceDiff, meta interface{}) bool {
+	reinstall, ok := d.GetOk("reinstall")
+
+	// Prior behaviour was to always destroy and create,
+	// so in the event we can't get the reinstall config; let's
+	// continue to do so.
+	if !ok {
+		return true
+	}
+
+	// We didn't get a reinstall configuration
+	reinstall_list, ok := reinstall.([]interface{})
+	if !ok {
+		return true
+	}
+
+	reinstall_config, ok := reinstall_list[0].(map[string]interface{})
+
+	// We didn't get a reinstall configuration
+	if !ok {
+		return true
+	}
+
+	return !reinstall_config["enabled"].(bool)
 }
 
 func resourceMetalDeviceCreate(d *schema.ResourceData, meta interface{}) error {
@@ -522,22 +526,10 @@ func resourceMetalDeviceCreate(d *schema.ResourceData, meta interface{}) error {
 
 	d.SetId(newDevice.ID)
 
-	// Wait for the device so we can get the networking attributes that show up after a while.
-	state, err := waitForDeviceAttribute(d, []string{"active", "failed"}, []string{"queued", "provisioning"}, "state", meta)
-	if err != nil {
-		d.SetId("")
-		fErr := friendlyError(err)
-		if isForbidden(fErr) {
-			// If the device doesn't get to the active state, we can't recover it from here.
+	if err = waitForActiveDevice(d, meta); err != nil {
+		return err
+	}
 
-			return errors.New("provisioning time limit exceeded; the Equinix Metal team will investigate")
-		}
-		return fErr
-	}
-	if state != "active" {
-		d.SetId("")
-		return fmt.Errorf("Device in non-active state \"%s\"", state)
-	}
 	/*
 			    Possibly wait for device network state
 		    	_, err := waitForDeviceAttribute(d, []string{"layer3"}, []string{"hybrid", "layer2-bonded", "layer2-individual"}, "network_type", meta)
@@ -716,6 +708,10 @@ func resourceMetalDeviceUpdate(d *schema.ResourceData, meta interface{}) error {
 		if _, err := client.Devices.Reinstall(d.Id(), &reinstallOptions); err != nil {
 			return friendlyError(err)
 		}
+
+		if err = waitForActiveDevice(d, meta); err != nil {
+			return err
+		}
 	}
 
 	return resourceMetalDeviceRead(d, meta)
@@ -778,5 +774,27 @@ func resourceMetalDeviceDelete(d *schema.ResourceData, meta interface{}) error {
 			}
 		}
 	}
+	return nil
+}
+
+func waitForActiveDevice(d *schema.ResourceData, meta interface{}) error {
+	// Wait for the device so we can get the networking attributes that show up after a while.
+	state, err := waitForDeviceAttribute(d, []string{"active", "failed"}, []string{"queued", "provisioning", "reinstalling"}, "state", meta)
+	if err != nil {
+		d.SetId("")
+		fErr := friendlyError(err)
+		if isForbidden(fErr) {
+			// If the device doesn't get to the active state, we can't recover it from here.
+
+			return errors.New("provisioning time limit exceeded; the Equinix Metal team will investigate")
+		}
+		return fErr
+	}
+
+	if state != "active" {
+		d.SetId("")
+		return fmt.Errorf("Device in non-active state \"%s\"", state)
+	}
+
 	return nil
 }
--- a/metal/resource_metal_device.go
+++ b/metal/resource_metal_device.go
@@ -1,6 +1,7 @@
 package metal
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -12,6 +13,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/errwrap"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/structure"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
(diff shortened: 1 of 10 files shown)


go.sum
```diff
@@ -253,8 +253,8 @@ github.com/nsf/jsondiff v0.0.0-20200515183724-f29ed568f4ce h1:RPclfga2SEJmgMmz2k
 github.com/nsf/jsondiff v0.0.0-20200515183724-f29ed568f4ce/go.mod h1:uFMI8w+ref4v2r9jz+c9i1IfIttS/OkmLfrk1jne5hs=
 github.com/oklog/run v1.0.0 h1:Ru7dDtJNOyC66gQ5dQmaCa0qIsAUFY3sFpK1Xk8igrw=
 github.com/oklog/run v1.0.0/go.mod h1:dlhp/R75TPv97u0XWUtDeV/lRKWPKSdTuV0TZvrmrQA=
-github.com/packethost/packngo v0.14.1 h1:hufjiwnWrxuI1iolRt/ZQw8E8YQeTDGIjB8vjELYGlY=
-github.com/packethost/packngo v0.14.1/go.mod h1:YrtUNN9IRjjqN6zK+cy2IYoi3EjHfoWTWxJkI1I1Vk0=
+github.com/packethost/packngo v0.17.0 h1:fGPlj9NDt6ejOrAMfUgx955oCaR1QBKA9pec14m0L38=
+github.com/packethost/packngo v0.17.0/go.mod h1:YrtUNN9IRjjqN6zK+cy2IYoi3EjHfoWTWxJkI1I1Vk0=
```
displague, line 257: This can be picked up in a rebase (we also vendor/, which the rebase will pick up)

rawkode: To confirm: You want me to remove the dependency changes for now?

metal/resource_metal_device.go
```diff
@@ -348,7 +350,61 @@ func resourceMetalDevice() *schema.Resource {
...
+							Description: "Whether the non-OS disks should be kept or wiped during reinstall",
+							Optional:    true,
+							Default:     false,
+						},
+					},
+				},
+			},
 		},
+		CustomizeDiff: customdiff.Sequence(
+			customdiff.ForceNewIf("reinstall", func(_ context.Context, d *schema.ResourceDiff, meta interface{}) bool {
```
displague, line 382: I haven't used ForceNewIf before, but if I understand this, it seems that we should supply each of these fields as the first argument: `userdata`, `customdata`, `operatingsystem`, and only return true if `reinstall` is not enabled.

Changes to the `reinstall` by itself should not prompt any immediate action or override any other diffs detected.

displague: With all three fields being processed as a ForceNewIf, I don't know if Sequence or All would provide a better experience (if not necessary).

https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/customdiff/compose.go#L11-L46

metal/resource_metal_device.go
```diff
@@ -649,9 +705,55 @@ func resourceMetalDeviceUpdate(d *schema.ResourceData, meta interface{}) error {
...
 	}
+
+	if d.HasChange("operating_system") || d.HasChange("user_data") || d.HasChange("custom_data") {
+		reinstallOptions, err := getReinstallOptions(d)
+
+		if err != nil {
+			return friendlyError(err)
+		}
+
+		if _, err := client.Devices.Reinstall(d.Id(), &reinstallOptions); err != nil {
```
displague, line 716: Terraform providers should not return a device until it is in a usable state.
We will have to do a wait after calling Reinstall (as we do after Create).

We may have to teach the wait loop that `reinstalling` is an expected transitional state.

displague: We may need a different Wait function if the existing one is not configurable (if we can't give it the desired state, and errors states).

We may also want to check the Update timeout limits to see if this change will cause Terraform to bump into those limits.  I can't think of any other long-running Update processes today (ignoring ForceNew).

metal/resource_metal_device.go
```diff
@@ -348,7 +350,61 @@ func resourceMetalDevice() *schema.Resource {
 				Default:     false,
 				ForceNew:    false,
 			},
+			"reinstall": {
```
displague, line 353: These fields will have to be documented in docs/resources/device.md.

We have some ongoing work to automate this in #13.

metal/resource_metal_device.go
rawkode: Fixed in new commit.

metal/resource_metal_device.go
```diff
@@ -348,7 +350,61 @@ func resourceMetalDevice() *schema.Resource {
...
+					Schema: map[string]*schema.Schema{
+						"enabled": {
+							Type:        schema.TypeBool,
+							Description: "Whether the device should be reinstalled instead of destroyed",
+							Optional:    true,
+							Default:     false,
+						},
+						"deprovision_fast": {
+							Type:        schema.TypeBool,
+							Description: "Whether the OS disk should written with 0 bytes",
```
displague, line 367:
```suggestion
							Description: "Whether the OS disk should be filled with `00h` bytes before reinstall",
```

metal/resource_metal_device.go
```diff
@@ -649,9 +705,55 @@ func resourceMetalDeviceUpdate(d *schema.ResourceData, meta interface{}) error {
...
+		return packngo.DeviceReinstallFields{}, fmt.Errorf("expected reinstall configuration and none available")
+	}
+
+	reinstall_config, ok := reinstall_list[0].(map[string]interface{})
+
+	if !ok {
+		return packngo.DeviceReinstallFields{}, fmt.Errorf("expected reinstall configuration and none available")
+	}
+
+	preserveData, ok := reinstall_config["preserve_data"].(bool)
```
displague, line 737: The field is called `persist_data` (not `preserve_data`).

The `if !ok { ... = false }` looks to be redundant, we would only need this condition to optionally set a `DeviceReinstallFields` field where the absence of the field (nil or omitted) would have a different meaning than `false`.

metal/resource_metal_device.go
```diff
@@ -649,9 +705,55 @@ func resourceMetalDeviceUpdate(d *schema.ResourceData, meta interface{}) error {
...
+	if !ok {
+		preserveData = false
+	}
+
+	deprovisionFast, ok := reinstall_config["preserve_data"].(bool)
```
displague, line 742: deprovisionFast
```suggestion
	deprovisionFast, ok := reinstall_config["deprovision_fast"].(bool)
```

go.sum
displague: `main` already has the dependency change. Just rebase and you will get it for free via #151.

The other dependency changes (the vendoring of new sdk customize .go files) seems important to keep in this PR.

metal/resource_metal_device.go
```diff
@@ -348,8 +350,66 @@ func resourceMetalDevice() *schema.Resource {
...
+							Description: "Whether the non-OS disks should be kept or wiped during reinstall",
+							Optional:    true,
+							Default:     false,
+						},
+					},
+				},
+			},
 		},
+		CustomizeDiff: customdiff.Sequence(
+			customdiff.ForceNewIf("custom_data", shouldReinstall),
```
displague, line 382: Nice condensing!

I wonder if `shouldReinstall` needs to take the field as an argument so it can check the that `custom_data` has infact changed before acting on it based on `reinstall`.

I don't know if `CustomizeDiff` is only triggered in the presence of a change in the named field or if it is always called, allowing the provider to use external factors to report a diff.

displague: func shouldReinstall(field string) func (_ context.Context, d *schema.ResourceDiff, meta interface{}) bool {
 return func (_ context.Context, d *schema.ResourceDiff, meta interface{}) bool {
    if ! diff on field return false
    ... do the existing function 
  }
}

rawkode: According to the docs:

> The return value of the condition function is ignored if the old and new values of the field compare equal, since no attribute diff is generated in that case.

https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/helper/customdiff?utm_source=godoc#ForceNewIf

rawkode:
Thanks for all that feedback, @displague . Tests running locally now, will let you know how they get on

metal/resource_metal_device.go
```diff
@@ -678,3 +776,25 @@ func resourceMetalDeviceDelete(d *schema.ResourceData, meta interface{}) error {
...
 	return nil
 }
+
+func waitForActiveDevice(d *schema.ResourceData, meta interface{}) error {
+	// Wait for the device so we can get the networking attributes that show up after a while.
+	state, err := waitForDeviceAttribute(d, []string{"active", "failed"}, []string{"queued", "provisioning", "reinstalling"}, "state", meta)
+	if err != nil {
+		d.SetId("")
+		fErr := friendlyError(err)
+		if isForbidden(fErr) {
```
displague, line 786: @t0mk We may want to rewrite this as a separate issue. I believe 404 is returned now for devices that have been moved to the failed provisioning project.

We can also offer more details than `the Equinix Metal team will investigate`, which may be misleading.

I'm thinking:
- return the error as-is wrapped with error text that we timed out waiting for active state
- include the instance id in the error
- if the error was 403/404, log that the Equinix Metal team will investigate (including a link to where this is stated in the docs) (don't say 'provisioning time limit exceeded', it could be sometime else, like invalid ipxe script, etc)

metal/resource_metal_device.go
```diff
@@ -348,8 +350,66 @@ func resourceMetalDevice() *schema.Resource {
...
+		return true
+	}
+
+	// We didn't get a reinstall configuration
+	reinstall_list, ok := reinstall.([]interface{})
+	if !ok {
+		return true
+	}
+
+	reinstall_config, ok := reinstall_list[0].(map[string]interface{})
```
displague, line 405: I'm sorry you had to experience this 😅 
Too bad the SDK doesn't provide `d.GetFirstMapStringOk()` (heh).

metal/resource_metal_device.go
```diff
@@ -649,9 +697,59 @@ func resourceMetalDeviceUpdate(d *schema.ResourceData, meta interface{}) error {
...
+
+	if !ok {
+		return packngo.DeviceReinstallFields{}, fmt.Errorf("expected reinstall configuration and none available")
+	}
+
+	preserveData, ok := reinstall_config["preserve_data"].(bool)
+	if !ok {
```
displague, line 738: The `!ok` blocks are redundant, the value here and for deprovisionFast will already be false.
We can `_` discard `ok`.

rawkode:
Tests seem to be passing 👍🏻

metal/resource_metal_device.go
displague: Created https://github.com/equinix/terraform-provider-metal/issues/153

displague (approved):
Thanks, @rawkode! This looks good. I'm going to merge it and let the full E2E tests run on this.
We'll address anything that comes up in a subsequent PR.  This new functionality seems well isolated from the rest of the typical workflow.

I'm interested in what new failure domains we may have. Other fields have been successfully updated but not reflected in state awaiting a successful reinstall. What happens to the device state when that reinstall fails (or doesn't return in time).  I think a `refresh`, part of the user's follow-up `apply` will handle any missed state.

You (as part of the org) should be able to push your branches directly to this org to get E2E tests run against your branches.

PR #156, opened 2021-07-15
fix: ensure go mod vendor is up to date

rawkode:
Currently the build is failing on the main branch, this should fix it.


rawkode:
Nevermind, I see @displague has also tackled this in #154

PR #157, opened 2021-07-15
Add resource and datasource `metal_gateway`, tests and docs

### LINKED ISSUE #138: Implement metal_gateway
The EM API will be delivering changes related to https://feedback.equinixmetal.com/networking-features/p/internet-gateway

This issue is a placeholder to reflect the packngo needed work to represent these API changes.

Depends on https://github.com/packethost/packngo/issues/288

t0mk:
This PR adds resource and datasource metal_gateway

fixes #138 

### DIFF
--- a/docs/data-sources/gateway.md
+++ b/docs/data-sources/gateway.md
@@ -1,0 +1,38 @@
+---
+page_title: "Equinix Metal: Metal Gateway"
+subcategory: ""
+description: |-
+  Retrieve Equinix Metal Gateways
+---
+
+# metal\_gateway
+
+Use this datasource to retrieve Metal Gateway resources in Equinix Metal.
+
+## Example Usage
+
+```hcl
+# Create Metal Gateway for a VLAN with a private IPv4 block with 8 IP addresses
+
+resource "metal_vlan" "test" {
+  description = "test VLAN in SV"
+  metro       = "sv"
+  project_id  = local.project_id
+}
+
+data "metal_gateway" "test" {
+  gateway_id               = local.gateway_id
+}
+```
+
+## Argument Reference
+
+* `gateway_id` - (Required) UUID of the metal gateway resource to retrieve
+
+## Attributes Reference
+
+* `project_id` - UUID of the project where the gateway is scoped to
+* `vlan_id` - UUID of the VLAN where the gateway is scoped to
+* `ip_reservation_id` - UUID of IP reservation block bound to the gateway
+* `private_ipv4_subnet_size` - Size of the private IPv4 subnet bound to this metal gateway, one of (8, 16, 32, 64, 128)`
+* `status` - Status of the gateway resource, "active" when ready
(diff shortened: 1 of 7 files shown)


t0mk:
@displague I have troubles testing this - the created Metal Gateways can't seem to move from state:activating to state:active.

The wait for state:active is timing out.

displague:
@t0mk we have an internal issue tracking this pre-release state transition problem.

displague:
The TF testing account did not have access to this feature. I believe it should be able to test on your next push. I don't believe the fix has been released for the `activating` state, yet.

t0mk:
@displague OK, I will make a bump commit once you confirm this was fixed.

displague:
We may need to hold off until next week. The `activating` state may be getting replaced with a `ready` state. 

Gateways with `ready` state will be ready for devices to be added to the assigned VLAN. 

The `active` will imply that any device is connected to the assigned VLAN.

Terraform will be able to release the created resource when it is in the `ready` state, which is what `POST` (2xx) will always return. The state may eventually transition to `active` (if any devices are on the VLAN), but it seems Terraform will not need to poll for state, release the gateway to the user immediately.  The 

There is a small window (measured in seconds) between `ready` and `active` where a server device connected to a VLAN won't be able to route traffic through it. Users should treat this like any temporary IP connectivity issue.

Interestingly, to avoid this, a user provision the gateway first and then use VLAN attachments (bulk or traditional) and await the assigned state to ensure that the device can take advantage of any gateways connected to the VLAN.

docs/resources/gateway.md
```diff
@@ -0,0 +1,61 @@
...
+  project_id = local.project_id
+  metro      = "sv"
+  quantity   = 2
+}
+
+resource "metal_gateway" "test" {
+  project_id        = local.project_id
+  vlan_id           = metal_vlan.test.id
+  ip_reservation_id = metal_reserved_ip_block.test.id
+}
```
displague, line 49: I think we would benefit from a guide that demonstrates devices (count = size of address space) being connected to the gateway, with userdata that configures the vlan and an address from the range.

Should we roll that in here as a guide, or another example, or stage this for later with a new issue, @t0mk? (I imagine you may already have tried this and may have the configuration handy).

t0mk: I would document the scerario in a guide. We can roll it in a separate PR I think.

t0mk:
@displague I dropped the state wait from Create functions, and fixes some other issues in tests and doc. This is ready for review.

docs/data-sources/gateway.md
```diff
@@ -0,0 +1,38 @@
...
+```hcl
+# Create Metal Gateway for a VLAN with a private IPv4 block with 8 IP addresses
+
+resource "metal_vlan" "test" {
+  description = "test VLAN in SV"
+  metro       = "sv"
+  project_id  = local.project_id
+}
+
+data "metal_gateway" "test" {
```
displague, line 23: We may want to offer a gateway_id way to lookup metal_reserved_ip_block datasources. 
https://registry.terraform.io/providers/equinix/metal/latest/docs/data-sources/reserved_ip_block

I'll open a new issue for that.

displague: It might also be helpful to lookup a metal_gateway by IP block or reservation. I'll open an issue for that.

displague:
Ran the tests locally and they passed - also enabled this account to perform the testing.

PR #169, opened 2021-07-26
Fix handling of metro attr in reserved IP block resource

### LINKED ISSUE #168: test failure from IP reservation that was not yet deleted
It appears that when deleting IP reservations, it may be necessary to wait for the reservation to be deleted before other operations continue.

```
=== RUN   TestAccMetalReservedIPBlock_Public
2021/07/23 13:41:32 [DEBUG] POST https://api.equinix.com/metal/v1/projects
2021/07/23 13:41:32 [DEBUG] GET https://api.equinix.com/metal/v1/projects/6287dc83-7a3e-44bc-9e19-b15f7261815f
2021/07/23 13:41:33 [DEBUG] GET https://api.equinix.com/metal/v1/projects/6287dc83-7a3e-44bc-9e19-b15f7261815f/bgp-config
2021/07/23 13:41:33 [DEBUG] POST https://api.equinix.com/metal/v1/projects/6287dc83-7a3e-44bc-9e19-b15f7261815f/ips
2021/07/23 13:41:34 [DEBUG] GET https://api.equinix.com/metal/v1/ips/11caa108-16fd-47a3-adc3-fc73bdd0f7cd
    resource_metal_reserved_ip_block_test.go:90: Step 1/1 error: After applying this test step, the plan was not empty.
        stdout:
        
        
        Terraform used the selected providers to generate the following execution
        plan. Resource actions are indicated with the following symbols:
        -/+ destroy and then create replacement
        
        Terraform will perform the following actions:
        
          # metal_reserved_ip_block.test must be replaced
        -/+ resource "metal_reserved_ip_block" "test" {
              ~ address        = "147.75.105.1" -> (known after apply)
              ~ address_family = 4 -> (known after apply)
              ~ cidr           = 31 -> (known after apply)
              ~ cidr_notation  = "147.75.105.0/31" -> (known after apply)
              ~ gateway        = "147.75.105.0" -> (known after apply)
              ~ global         = false -> (known after apply)
              ~ id             = "11caa108-16fd-47a3-adc3-fc73bdd0f7cd" -> (known after apply)
              ~ manageable     = true -> (known after apply)
              ~ management     = false -> (known after apply)
              - metro          = "ny" -> null # forces replacement
              ~ netmask        = "255.255.255.254" -> (known after apply)
              ~ network        = "147.75.105.0" -> (known after apply)
              ~ public         = true -> (known after apply)
                tags           = [
                    "Tag1",
                    "Tag2",
                ]
                # (4 unchanged attributes hidden)
            }
        
        Plan: 1 to add, 0 to change, 1 to destroy.
2021/07/23 13:41:35 [DEBUG] DELETE https://api.equinix.com/metal/v1/ips/11caa108-16fd-47a3-adc3-fc73bdd0f7cd
2021/07/23 13:41:35 [DEBUG] DELETE https://api.equinix.com/metal/v1/projects/6287dc83-7a3e-44bc-9e19-b15f7261815f
2021/07/23 13:41:35 [DEBUG] GET https://api.equinix.com/metal/v1/ips/11caa108-16fd-47a3-adc3-fc73bdd0f7cd
--- FAIL: TestAccMetalReservedIPBlock_Public (5.27s)
```

t0mk:
Fixes #168 

This PR adds logic to supress diff for metro attirbute for `metal_reserved_ip_block` if it's created in facility, e.g.:

```
resource "metal_reserved_ip_block" "test" {
	project_id  = metal_project.foobar.id
	facility    = "ewr1"
	type        = "public_ipv4"
	quantity    = 2
}
```

That block will get metro attr set from TF state, and will emit a diff because the metro is unset in HCL ("ny"->null). We want to suppress that diff. There is another suppressing check, I think because of migration. It's a bit messy, but it will most likely do the right thing - when metro is set to "ny" and then user changes it to "sv", the resource will be recreated.

There is a small issue coming with the diff suppress checks - if user creates the resource above, and then unsets the facility, Terraform will ignore it, even though it's an invalid conf (ip reservation must have facility or metro set). But I think it's not too much of a problem.

### DIFF
--- a/metal/resource_metal_reserved_ip_block.go
+++ b/metal/resource_metal_reserved_ip_block.go
@@ -100,17 +100,26 @@ func resourceMetalReservedIPBlock() *schema.Resource {
 		ForceNew:      true,
 		ConflictsWith: []string{"facility"},
 		Description:   "Metro where to allocate the public IP address block, makes sense only for type==public_ipv4, must be empty for type==global_ipv4, conflicts with facility",
-		DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+		DiffSuppressFunc: func(k, fromState, fromHCL string, d *schema.ResourceData) bool {
 			_, facOk := d.GetOk("facility")
-			// new - new val from template
-			// old - old val from state
-			//
-			// suppress diff if metro is manually set for first time, and
-			// facility is already set
-			if len(new) > 0 && old == "" && facOk {
-				return facOk
+
+			// if facility is not in state, treat the diff normally, otherwise do following messy checks:
+			if facOk {
+				// If metro from HCL is specified, but not present in state, supress the diff.
+				// This is legacy, and I think it's here because of migration, so that old
+				// facility reservations are not recreated when metro is specified ???)
+				if fromHCL != "" && fromState == "" {
+					return true
+				}
+				// If metro is present in state but not present in HCL, supress the diff.
+				// This is for "facility-specified" reservation blocks created after ~July 2021.
+				// These blocks will have metro "computed" to the TF state, and we don't want to
+				// emit a diff if the metro field is empty in HCL.
+				if fromHCL == "" && fromState != "" {
+					return true
+				}
 			}
-			return old == new
+			return fromState == fromHCL
 		},
 		StateFunc: toLower,
 	}
--- a/metal/resource_metal_reserved_ip_block_test.go
+++ b/metal/resource_metal_reserved_ip_block_test.go
@@ -17,10 +17,10 @@ resource "metal_project" "foobar" {
 }
 
 resource "metal_reserved_ip_block" "test" {
-	project_id = "${metal_project.foobar.id}"
-	type     = "global_ipv4"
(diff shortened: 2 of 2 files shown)


t0mk:
@displague please review. I tried to explain what's going on in the comments.

displague:
Let's add an extra step (if needed) to one of the existing IP Reservation tests to verify that the initial metro value (when facility was explicit) does not create a problem, and then let's add another test that replaces `facility` with the matching `metro` - this should not trigger a diff either.

> ip reservation must have facility or metro set

Unless the type is one of the global types.

t0mk:
> Let's add an extra step (if needed) to one of the existing IP Reservation tests to verify that the initial metro value (when facility was explicit) does not create a problem

The TestAccMetalReservedIPBlock_Public test is checking that, it was the reason the test was failing.

I will add a test creating a facility reservation and then changing it to matching metro.

>> ip reservation must have facility or metro set

> Unless the type is one of the global types.

Yes, I meant the condition for the public_ipv4 type. The resource will err if you will try to create a global reservation with metro or facility specified.

PR #170, opened 2021-07-27
Fixes in acceptance tests

t0mk:
This PR hopefully fixes some errors in acceptance tests.

### DIFF
--- a/metal/datasource_metal_project_ssh_key_test.go
+++ b/metal/datasource_metal_project_ssh_key_test.go
@@ -76,7 +76,9 @@ func TestAccMetalProjectSSHKeyDataSource_ByID(t *testing.T) {
 					resource.TestCheckResourceAttr(
 						datasourceName, "public_key", publicKeyMaterial),
 				),
-				ExpectNonEmptyPlan: true,
+				// Why was follwing flag set? The plan is applied and then it's empty.
+				// It's causing errors in acceptance tests. Was this because of some API bug?
+				//ExpectNonEmptyPlan: true,
 			},
 			{
 				Config:      testAccMetalProjectSSHKeyDataSourceConfig_noKey(keyName, publicKeyMaterial),
--- a/metal/resource_metal_spot_market_request_test.go
+++ b/metal/resource_metal_spot_market_request_test.go
@@ -26,9 +26,9 @@ data "metal_spot_market_request" "dreq" {
 }
 
 resource "metal_spot_market_request" "request" {
-  project_id       = "${metal_project.test.id}"
+  project_id       = metal_project.test.id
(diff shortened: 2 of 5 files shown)


t0mk:
Let's see (and measure) if there's an improvement over the test run in https://github.com/equinix/terraform-provider-metal/pull/169/checks?check_run_id=3171054901

This is former acc test log: [formerlog.txt](https://github.com/equinix/terraform-provider-metal/files/6885885/formerlog.txt), it contains 18 lines with string error.

t0mk:
Now there's just timeout for spot market request test and HTTP 500 error for TestAccMetalPortVlanAttachment_L2Native, so this is an improvement.

metal/resource_metal_spot_market_request_test.go
```diff
@@ -26,9 +26,9 @@ data "metal_spot_market_request" "dreq" {
 }
 
 resource "metal_spot_market_request" "request" {
-  project_id       = "${metal_project.test.id}"
+  project_id       = metal_project.test.id
   max_bid_price    = data.metal_spot_market_price.test.price * 1.2
-  facilities       = ["ewr1"]
+  facilities       = ["sv15"]
```
displague, line 31: In the future, maybe we can include data sources with these tests to determine which facility will be the best for the test: https://github.com/equinix/terraform-provider-metal/issues/164#issuecomment-884896686.  This would make testing dependent on an additional resource and API call, but it should help pass more tests overall.

(or perhaps we can make a single capacity API check at the start of the tests and use that facility in all tests? )

metal/datasource_metal_project_ssh_key_test.go
```diff
@@ -76,7 +76,9 @@ func TestAccMetalProjectSSHKeyDataSource_ByID(t *testing.T) {
 					resource.TestCheckResourceAttr(
 						datasourceName, "public_key", publicKeyMaterial),
 				),
-				ExpectNonEmptyPlan: true,
+				// Why was follwing flag set? The plan is applied and then it's empty.
```
displague, line 79: This may have been left over from experiments before finding https://github.com/packethost/terraform-provider-packet/pull/250#discussion_r490700522

displague:
We've been hitting timeouts in the acceptance tests. I noticed that TestAccMetalVLAN_* is not triggered.  Perhaps we should add 30 minutes.

metal/datasource_metal_project_ssh_key_test.go
displague: Removing the ExpectNonEmptyPlan and the comment seems good to me. We might want a comment explanation for the empty step that follows (maybe just a link to the terraform-provider-packet PR250 comment, or the hashicorp PR).

Issue #171, opened 2021-07-28
github action CI reports are too hard to read


t0mk:
The move from HashiCorp Teamcity to Github Actions worsened the quality of CI in this provider. 

I can't easily see:
- which tests have failed, in one screen
- which of these fails are new, in one screen
- which tests don't fail anymore, in one screen
- I can't unfold test fail and see stack trace of the error

This^ is not too much to ask form a CI tool in 2021. 

The console output of the Github actions is simply horrible. It's been made too hard to briefly check if a (small) change broke something in the other parts of the code. Engs will not have the time and patience to grep for "error" in thousands of lines in a web ui, and then scroll up several screens to find out which test have failed. This will negatively impact the quality of the provider - bug that could be easily spotted will go unnoticed. We run a huge amount of API traffic, and create and verify a huge amount of resources, it's a pity that the result of this effort is so hard to interpret.

displague:
Can we remove the `DEBUG` URL request lines? These create a lot of noise and generally do not need to be shown on passing tests (these may be helpful for failed tests).

displague:
It is possible to download the log and access logs via the `gh` cli: https://docs.github.com/en/actions/managing-workflow-runs/using-workflow-run-logs

displague:
A GH Action like https://github.com/marketplace/actions/golang-test-annotations could leave annotations on the failing tests which should make failures highly visible from the PR view.

displague:
It looks like GitHub has a feature in the works that will fix the error reporting for reports not directly included in the PR:
![image](https://user-images.githubusercontent.com/317653/146196114-95580811-1e5f-4121-a50e-3a30afcc51c5.png)

PR #173, opened 2021-08-04
Debug mode, Test failure annotations, Code Coverage

displague:
* Adds:
  * Debug mode (https://www.terraform.io/docs/extend/debugging.html#enabling-debugging-in-a-provider)
  * ~Test failure annotations (https://github.com/marketplace/actions/golang-test-annotations) #171~
  * Code Coverage (https://app.codecov.io/gh/equinix/terraform-provider-metal)
* Increases test duration to 180m due to recent timeouts.

### DIFF
--- a/.github/workflows/acctest.yml
+++ b/.github/workflows/acctest.yml
@@ -58,7 +58,7 @@ jobs:
         go mod download
 
     - name: TF acceptance tests
-      timeout-minutes: 120
+      timeout-minutes: 180
       env:
         TF_ACC: "1"
         TF_ACC_TERRAFORM_VERSION: ${{ matrix.terraform }}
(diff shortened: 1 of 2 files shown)


displague:
One drawback of `-json` and outputting the results directly to a file is that we can't monitor the progress of tests.  However, monitoring the progress of tests is not a great experience today because GHA doesn't reveal the output lines that were missed before loading the page until the job completes.

If we were to pipe to `tee`, we could see the output but in the `-json` format, they may be less readable.

.github/workflows/acctest.yml
```diff
@@ -68,10 +68,25 @@ jobs:
 
         METAL_AUTH_TOKEN: ${{ secrets.PACKET_AUTH_TOKEN }}
       run: |
-        go test -v -cover -parallel 4 -timeout 120m ./metal
+        go test -v -json -coverprofile coverage.txt -covermode=atomic -parallel 4 -timeout 180m ./metal > test.json
```
displague, line 71: Do coverage.txt and test.json have to be unique names for the TF 0.13 test and 1.0?

displague: Can we increase the parallelization?

.github/workflows/acctest.yml
```diff
@@ -68,10 +68,25 @@ jobs:
...
+        test-results: test.json
     - name: Sweeper
       if: ${{ always() }}
       env:
         METAL_AUTH_TOKEN: ${{ secrets.PACKET_AUTH_TOKEN }}
       run: |
-        go test ./metal -v -sweep="tf_test"
+        go test ./metal -v -json -sweep="tf_test"  > sweep.json
+    - name: Annotate sweep
+      if: always()
```
displague, line 84: This is a different format than the existing `${{ always() }}`.  I assume they are both acceptable.

.github/workflows/acctest.yml
```diff
@@ -68,10 +68,25 @@ jobs:
...
+      uses: guyarb/golang-test-annotations@v0.4.0
+      with:
+        test-results: sweep.json
+    - name: Upload coverage to Codecov
+      uses: codecov/codecov-action@v2
+      with:
+        token: ${{ secrets.CODECOV_TOKEN }}
```
displague, line 91: This token has been added to the repo settings.

displague:
Hmm, none of the errors were annotated.

displague:
> Hmm, none of the errors were annotated.

Annotation is only possible when the Test itself was changed.  The code to annotate must appear on the PR.

There are other GH actions available that would provide a summary of the test run and we may want to look into those.

t0mk (approved):
The debug mode code looks OK, although I havent' really tried to run the provide with delve. I've never used it and I'm not sure how I would go about it, but maybe it's a good alternative to debug prints (whihc are hard to do in a tf provider).

I tried to produce the coverage file and it created it alright. I didn't try to upload it to codecov.

Overall LGTM.

displague:
We should start seeing codecov posting on PRs - might need to activate the [webhook](https://app.codecov.io/gh/equinix/terraform-provider-metal/settings/general) if not.

[codecov](https://codecov.io/gh/equinix/terraform-provider-metal)

PR #179, opened 2021-08-30
Add desired_features attribute to facility datasource

### LINKED ISSUE #88: metal_facility data source should support filtering by feature flags
Module authors that want to enforce facility compatibility with their modules may benefit from feature flag filtering in the facility data source.

For example:

```
data "metal_facility" "hybrid_bonded_check" {
    code = var.facility
    features = ["ibx"] // all features selected must be present to match the facility
}
```

This resource would catch a mis-configuration created by choosing an invalid facility for a configuration that requires hybrid-bonded interfaces.  The error message would not be very helpful, but the configuration would fail much earlier.

t0mk:
This PR adds `desired_features` attribute to facility datasource, so that users can ensure that selected facility has certain capabilities.

Fixes #88 

### DIFF
--- a/metal/datasource_metal_facility.go
+++ b/metal/datasource_metal_facility.go
@@ -51,7 +51,7 @@ func dataSourceMetalFacility() *schema.Resource {
 				Description: "The code of the Facility to match",
 				Required:    true,
 			},
-			"desired_features": {
+			"features_required": {
 				Type:        schema.TypeSet,
 				Description: "Features which the facility needs to have.",
 				Elem:        &schema.Schema{Type: schema.TypeString},
@@ -108,7 +108,7 @@ func dataSourceMetalFacilityRead(d *schema.ResourceData, meta interface{}) error
 		return fmt.Errorf("Error listing Facilities: %s", err)
 	}
 
-	dfRaw, dfOk := d.GetOk("desired_features")
+	dfRaw, dfOk := d.GetOk("features_required")
 
 	for _, f := range facilities {
 		if f.Code == code {
--- a/metal/datasource_metal_facility.go
+++ b/metal/datasource_metal_facility.go
@@ -51,6 +51,13 @@ func dataSourceMetalFacility() *schema.Resource {
 				Description: "The code of the Facility to match",
 				Required:    true,
 			},
+			"desired_features": {
+				Type:        schema.TypeSet,
+				Description: "Features which the facility needs to have.",
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Optional:    true,
+				MinItems:    1,
+			},
 			"name": {
 				Type:        schema.TypeString,
 				Description: "The name of this Facility.",
@@ -101,8 +108,16 @@ func dataSourceMetalFacilityRead(d *schema.ResourceData, meta interface{}) error
 		return fmt.Errorf("Error listing Facilities: %s", err)
 	}
 
+	dfRaw, dfOk := d.GetOk("desired_features")
+
 	for _, f := range facilities {
 		if f.Code == code {
+			if dfOk {
+				unsupported := difference(convertStringArr(dfRaw.(*schema.Set).List()), f.Features)
+				if len(unsupported) > 0 {
+					return fmt.Errorf("facililty %s doesn't have feature(s) %v", f.Code, unsupported)
+				}
+			}
 			d.SetId(f.ID)
 			return setMap(d, map[string]interface{}{
 				"code":     f.Code,
--- a/metal/datasource_metal_facility_test.go
+++ b/metal/datasource_metal_facility_test.go
@@ -63,7 +63,7 @@ func testAccDataSourceFacilityConfigFeatures() string {
 	return `
 data "metal_facility" "test" {
     code = "ewr1"
-    desired_features = ["ibx"]
+    features_required = ["baremetal", "ibx"]
 }
 `
 }
--- a/metal/datasource_metal_facility_test.go
+++ b/metal/datasource_metal_facility_test.go
@@ -8,8 +8,14 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 )
 
+var (
+	matchErrMissingFeature = regexp.MustCompile(`.*doesn't have feature.*`)
+	matchErrNoCapacity     = regexp.MustCompile(`Not enough capacity.*`)
+)
+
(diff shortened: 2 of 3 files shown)


metal/datasource_metal_facility.go
```diff
@@ -17,6 +17,13 @@ func dataSourceMetalFacility() *schema.Resource {
 				Description: "The code of the Facility to match",
 				Required:    true,
 			},
+			"desired_features": {
```
displague, line 20: My first thought was that we should call this field `features`, but then I see that we have the existing `features` attribute. Hmm.

Looking at the description, "which the facility **needs to have**." `required_features` maybe?

What are your thoughts on using `features` with optional + computed?
I've done this before here: https://registry.terraform.io/providers/linode/linode/latest/docs/data-sources/domain_record (and in domain).

I keep running into examples on DO, Linode, and AWS with a `filter { name = "" values = "" }` pattern. We can implement that for arbitrary server-side filtering (when the API supports it). I've opened https://github.com/equinix/terraform-provider-metal/issues/180 to address this.  

When a field is not API filterable, the Terraform argument field name tends to either match the attribute field name or include the name of the filtering operation in the name: `"name_regex" ... "uses local filtering, not available in the API, to allow for more flexible filtering"`.  In this case, the name might be `features_include` or `features_required`.

metal/datasource_metal_facility.go
```diff
@@ -50,8 +57,16 @@ func dataSourceMetalFacilityRead(d *schema.ResourceData, meta interface{}) error
...
 	}
 
+	dfRaw, dfOk := d.GetOk("desired_features")
+
 	for _, f := range facilities {
 		if f.Code == code {
+			if dfOk {
+				unsupported := difference(convertStringArr(dfRaw.(*schema.Set).List()), f.Features)
+				if len(unsupported) > 0 {
+					return fmt.Errorf("facililty %s doesn't have feature(s) %v", f.Code, unsupported)
```
displague, line 67: Shouldn't we `continue` until we find a match?  Only error if there were no matches?

The logic has to be changed since we are no longer accepting the first match.

metal/datasource_metal_facility_test.go
```diff
@@ -22,10 +23,35 @@ func TestAccFacilityDataSource_Basic(t *testing.T) {
...
+			},
+		},
+	})
+}
+
+func testAccDataSourceCheckMetalFacilityConfigFeatures() string {
+	return `
+data "metal_facility" "test" {
+    code = "ewr1"
+    desired_features = ["ibx"]
```
displague, line 54: This will be a very useful feature!

displague: I think it will be more practical to use `metro` instead of `code` or `country` in this example.

"I want any *IBX* featured facility in *{metro or address component (country, state, etc)}*" 
https://github.com/packethost/packngo/blob/v0.19.0/facilities.go#L15

We don't have client-side filters for those fields, however. Perhaps we can introduce that in a follow-up PR? (or perhaps server-side filtering supports this?)

metal/datasource_metal_facility.go
t0mk: `features_required` sounds good.

> What are your thoughts on using features with optional + computed?

I think it would be messy. It makes the Computed aspect of the field really hard to use. There's no Copmuted + Optional field in the domain record datasource:
https://github.com/linode/terraform-provider-linode/blob/main/linode/data_source_linode_domain_record.go

metal/datasource_metal_facility.go
t0mk: The code is `Required`. If the code matches, then this is the fac we're looking for.

This PR doesn't return a list (or first found), it returns just one facility. It doesn't really filter, just checks that the found fac has the properties.

metal/datasource_metal_facility.go
displague: That's interesting. `name` doesn't say optional + computed, but it is treated that way:

https://github.com/linode/terraform-provider-linode/blob/main/linode/data_source_linode_domain_record.go#L88-L117

d.Set is not error-checked there, I suppose this could be erroring.

metal/datasource_metal_facility.go
displague: I see. Having a code in mind reduces the usability of this new feature to me (I'm thinking about this as https://github.com/equinix/terraform-provider-metal/issues/180#issuecomment-909863036). 

This feature will still be helpful when validating that some user-supplied `var.facility` used in a module or configuration is a facility with the correct features.

metal/datasource_metal_facility_test.go
displague: I wasn't thinking about the required field, `code`.

This makes sense.

Maybe someday this will be possible (https://github.com/equinix/terraform-provider-metal/pull/179/files/6b8eaf81c5b1f118b2305f5d5e54ef218e774e9d#r699831442).

t0mk:
@displague I changed `desired_features` to `features_required`.

metal/datasource_metal_facility.go
```diff
@@ -17,6 +17,13 @@ func dataSourceMetalFacility() *schema.Resource {
 				Description: "The code of the Facility to match",
 				Required:    true,
 			},
+			"features_required": {
+				Type:        schema.TypeSet,
+				Description: "Features which the facility needs to have.",
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Optional:    true,
+				MinItems:    1,
```
displague, line 25: Since this field is optional, I think `MinItems: 0` is more fitting (https://github.com/equinix/terraform-provider-metal/pull/179/files#diff-df84d3f7d9e865eacdfcba301894d3db8d7c40830f29e13cd7cf02b585eb8a6bR65 looks like it will work as-is). I can imagine a configuration that takes user variables, perhaps in a map, where each server or facility has different feature requirements. Some of those servers or facilities may not need any features while others do.  

It would be more convenient to do:

```
data "metal_facility" "test" {
  count = len(var.configs)

  code = var.configs[count.index].facility.code
  features_required = var.configs[count.index].facility.features
}
```

than:

```
data "metal_facility" "test" {
  count = len(var.configs)

  code = var.configs[count.index].facility.code
  features_required = len(var.configs[count.index].facility.features) > 0 ? var.configs[count.index].facility.features : null
}
```

Although the user could just as easily put `null` in their `var.configs[].facility.features`.

displague: We can revisit this later.

PR #181, opened 2021-08-30
Capacity check for facility and metro datasources

t0mk:
This PR adds capacity field to facility datasource in order to verify that a particular facility has enough capacity. See usage in the test.

_Originally https://github.com/equinix/terraform-provider-metal/issues/88#issuecomment-892999121_

### DIFF
--- a/metal/datasource_metal_facility.go
+++ b/metal/datasource_metal_facility.go
@@ -33,6 +33,26 @@ func dataSourceMetalFacility() *schema.Resource {
 				Elem:        &schema.Schema{Type: schema.TypeString},
 				Computed:    true,
 			},
+			"capacity": {
+				Type:        schema.TypeList,
+				Description: "Optional capacity specification",
+				MaxItems:    1,
+				Optional:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"plan": {
+							Type:        schema.TypeString,
+							Description: "Plan which has to be available in selected facility",
+							Required:    true,
+						},
+						"quantity": {
+							Type:     schema.TypeInt,
+							Default:  1,
+							Optional: true,
+						},
+					},
+				},
+			},
 		},
 	}
 }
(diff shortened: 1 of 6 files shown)


displague:
This filter seems like a good blocker to ensure that the chosen facility meets the expected criteria.

I would like to continue to explore what it looks like to automatically choose a plan, facility, and metro based on capacity.

Hypothetical:

```
data "metal_plan" "smallest" {
  class = "baremetal" # or is this 'line'?
  plan_in = ["t1.small.x86", "c3.small.x86"]
  level = "normal"
  metro_in = ["dc", "ny", "da"]
}

data "metal_facility" "ibx" {
  metro_in = metal_plan.smallest.metros # array
  features_required = ["ibx"]
  # optional code, in this case
} 

resource "metal_device" "foo" {
  facility = data.metal_facility.ibx.code
  plan = data.metal_plan.smallest.code
   ...
}
```

metal/datasource_metal_facility.go
```diff
@@ -33,6 +33,26 @@ func dataSourceMetalFacility() *schema.Resource {
 				Elem:        &schema.Schema{Type: schema.TypeString},
 				Computed:    true,
 			},
+			"capacity": {
+				Type:        schema.TypeList,
+				Description: "Optional capacity specification",
+				MaxItems:    1,
```
displague, line 39: Should we allow for multiple capacity blocks?  What if I want a facility with 2 n2 devices and 3 c3 devices.

metal/datasource_metal_facility.go
```diff
@@ -33,6 +33,26 @@ func dataSourceMetalFacility() *schema.Resource {
...
+				Optional:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"plan": {
+							Type:        schema.TypeString,
+							Description: "Plan which has to be available in selected facility",
+							Required:    true,
+						},
+						"quantity": {
```
displague, line 48: We don't have to introduce the `level` checks in this PR, but we may want to add them in the future.  

For users that know which facility they want and would like to prevent provisioning if their `var.device_count` won't be met, this will be helpful.

metal/datasource_metal_facility.go
displague: I suppose we could think about that later and users could just use multiple `data "metal_facility"` blocks for now. (since this resource conditionally represents a `code` predefined facility)

t0mk: It's actually not too hard to do it right now, the response on a capacity request looks following:

```
{"servers":[
  {"facility":"sv15","plan":"c3.small.x86","quantity":1,"available":true},
  {"facility":"sv15","plan":"c3.medium.x86","quantity":3,"available":false}
]}
```

We can agree that the datasource will throw an error if any of the sections has `available:false`. The error will actually say which capacity requirement can't be satisfied.

metal/datasource_metal_facility.go
t0mk: - I think level is not part of the capacity query: https://metal.equinix.com/developers/api/capacity/#check-capacity
- I think the levels are useless once you can actually query if a faciltiy (or metro) has *n* devices of a particular plan available.

t0mk:
@displague I made it possible to use more than one capacity block. I also added the same funtionality to the metro datasoruce. Please take a look, this is ready for review.

displague:
I ran the acceptance tests locally and they passed.

PR #182, opened 2021-09-03
Add termination_time attr for better sweeping of test devices

t0mk:
This PR adds `termination_time` attirbute to metal_device". We can use it to make sure devices self-destruct in acceptance tests.

It should probably not be used by normal users as terraform apply run will attempt to create a tracked device after it self-destructed.

towards #178

### DIFF
--- a/metal/resource_metal_device.go
+++ b/metal/resource_metal_device.go
@@ -355,10 +355,9 @@ func resourceMetalDevice() *schema.Resource {
 			},
 			"termination_time": {
 				Type:        schema.TypeString,
-				Description: "Timestamp for device termination, use with care. For example \"2021-09-03T16:32:00+03:00\". If you don't supply timezone info, timestamp is assumed to be in UTC.",
+				Description: "Timestamp for device termination. For example \"2021-09-03T16:32:00+03:00\". If you don't supply timezone info, timestamp is assumed to be in UTC.",
 				Optional:    true,
 				ForceNew:    false,
-				Deprecated:  "You probably don't want to use this attribute, we added it for sweeping test artifacts.",
 				ValidateFunc: func(val interface{}, key string) (warns []string, errs []error) {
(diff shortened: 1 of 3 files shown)


t0mk:
@displague I should have made this a draft. Can you suggest which tests I should add the termination time to? We have the L2 acceptance tests timing out. No problem if you can't take a look, I will do it on Monday then.

Please consider this still a draft PR.

metal/resource_metal_device.go
```diff
@@ -353,6 +353,21 @@ func resourceMetalDevice() *schema.Resource {
 				Default:     false,
 				ForceNew:    false,
 			},
+			"termination_time": {
+				Type:        schema.TypeString,
+				Description: "Timestamp for device termination, use with care. For example \"2021-09-03T16:32:00+03:00\". If you don't supply timezone info, timestamp is assumed to be in UTC.",
+				Optional:    true,
+				ForceNew:    false,
+				Deprecated:  "You probably don't want to use this attribute, we added it for sweeping test artifacts.",
```
displague, line 361: I don't think we need to be defensive or prescriptive about this field.  The API exposes it so users may consume it however they like.

For example:

```hcl
variable "temporary_hours" { default = 14*24 } 
variable "expires" { default = timeadd(formatdate("YYYY-MM-01T00:00:00-00:00"), "${var.temp_hours}h") }
variable "temp_count" { default = 3 }

// add 3 temporary servers for extra work needed at the beginning of the month
resource "metal_device" "temporary" {
  count = var.expires > timestamp() ? var.temp_count : 0
  termination_time =  var.expires
}
```

t0mk: Alright!

t0mk:
@displague should I add the termination_time to metal_devices in test in here, or in a separate PR? I'm thinking separate PR, because we'll also need to decide the deadline, how to pass it to the tests (a lot of string formatting for the configs), ans maybe some other issues.

PR #189, opened 2021-09-10
fix Virtual Circuit disconnecting before delete

### LINKED ISSUE #183: Deletion of Metal_Virtual_Connection fails with Error 422
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
Run `terraform -v` to show the version. If you are not running the latest version of Terraform, please upgrade because your issue may have already been fixed.

Affected Resource(s)
Please list the resources as a list, for example:
- metal_virtual_circuit

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

Terraform Configuration Files
```hcl

resource "metal_virtual_circuit" "vc1" {
    connection_id = var.conn_id
    project_id = var.project_id
    port_id = var.port_id
    vlan_id = metal_vlan.vlan1.id
    nni_vlan = var.nni_vlan
}
```

### Debug Output
Get this error message then using equinix metal provider version >= 3.0. When provider version is reverted to 2.1.0, no error messages and virtual connection is deleted as expected. 

metal_virtual_circuit.vc1: Destroying... [id=31e13e9e-57e8-4720-a9b2-a715106a9a8c]
╷
│ Error: PUT https://api.equinix.com/metal/v1/virtual-circuits/31e13e9e-57e8-4720-a9b2-a715106a9a8c: 422 Update params can't be blank 

### Expected Behavior
Virtual Connection gets deleted when `terraform destroy` is run.

### Actual Behavior
Get the error message above when running provider version >=3.0.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply` - Everything gets applied as expected
2. `terraform destroy`- action errors out with the warning above.

### Important Factoids
Are there anything atypical about your accounts that we should know?
no

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:
no

[terraform.txt](https://github.com/equinix/terraform-provider-metal/files/7108059/terraform.txt)

displague:
_I suspect this..._
Fixes #183 

`Error: PUT https://api.equinix.com/metal/v1/virtual-circuits/31e13e9e-57e8-4720-a9b2-a715106a9a8c: 422 Update params can't be blank`

https://github.com/packethost/packngo/blob/86fc114a1bf21d9f0ad2f16ddbbecd769b79ee4a/virtualcircuits.go#L40 
The `omitempty` here will prevent the `nil` `VirtualNetworkID` from being sent to the API. By using an empty string, the API will receive something, however it is not clear to me if the API will accept `""` to declare the intention to disconnect the VLAN, or if we will need to change packngo to permit `null` to be sent.  In this case, we could use an option that may be undefined/omitted or literal null or string:
```
VirtualNetworkID **string `json:"vnid,omitempty"`, 
```

### DIFF
--- a/metal/resource_metal_virtual_circuit.go
+++ b/metal/resource_metal_virtual_circuit.go
@@ -249,9 +249,10 @@ func resourceMetalVirtualCircuitUpdate(d *schema.ResourceData, meta interface{})
 func resourceMetalVirtualCircuitDelete(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*packngo.Client)
 	// we first need to disconnect VLAN from the VC
+	empty := ""
 	_, _, err := client.VirtualCircuits.Update(
 		d.Id(),
-		&packngo.VCUpdateRequest{VirtualNetworkID: nil},
+		&packngo.VCUpdateRequest{VirtualNetworkID: &empty},
 		nil,
 	)
 	if err != nil {


t0mk:
@displague looks like a good temporary solution. I should have seen that the `omitempty` will not work as expected iwth the nil there.

Would be good to know what exactly the API expects there for the removal. It should be possible to observe from web UI console (with debugging tools) if you have a connection for testing available.

displague:
We may not need to remove the VLAN explicitly. When I can test this I will try taking that approach too.

PR #191, opened 2021-09-13
Add `end_at` param to spot_market_request to allow limiting SMR duration

t0mk:
This PR adds `end_at` parameter to metal_spot_market_request to allow removing the SMR in future. Will be useful for our acceptance tests.

### DIFF
--- a/docs/resources/spot_market_request.md
+++ b/docs/resources/spot_market_request.md
@@ -39,6 +39,7 @@ The following arguments are supported:
 * `max_bid_price` - (Required) Maximum price user is willing to pay per hour per device
 * `project_id` - (Required) Project ID
 * `wait_for_devices` - (Optional) On resource creation - wait until all desired devices are active, on resource destruction - wait until devices are removed
+* `end_at` - (Optional) Deadline for the request. For example \"2021-09-03T16:32:00+03:00\". If you don't supply timezone info, timestamp is assumed to be in UTC."
 * `facilities` - (Optional) Facility IDs where devices should be created
 * `metro` - (Optional) Metro where devices should be created
 * `locked` - (Optional) Blocks deletion of the SpotMarketRequest device until the lock is disabled
@@ -47,7 +48,6 @@ The following arguments are supported:
   * `plan`
   * `operating_system`
   * `hostname`
-  * `termintation_time`
   * `always_pxe`
   * `description`
   * `features`
--- a/metal/provider_test.go
+++ b/metal/provider_test.go
@@ -45,3 +45,7 @@ func testAccPreCheck(t *testing.T) {
 func testDeviceTerminationTime() string {
 	return time.Now().UTC().Add(60 * time.Minute).Format(time.RFC3339)
 }
+
+func testSpotMarketRequestTerminationTime() string {
+	return time.Now().UTC().Add(60 * time.Minute).Format(time.RFC3339)
+}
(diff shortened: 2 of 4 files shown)


docs/resources/spot_market_request.md
```diff
@@ -47,7 +48,6 @@ The following arguments are supported:
   * `plan`
   * `operating_system`
   * `hostname`
-  * `termintation_time`
```
displague, line 50: termination_time is available as an instance_parameter attribute.
Perhaps this should move to the attributes section?

displague: The test could confirm that termination_time attribute is set too.

t0mk: @displague the `termination_time` in the instance params is a problem of itself:

- there is a typo in the argument name (it's `termin>t<ation_time` also in the code)
- I don't think it actually works, there would have to be some special handling in the Read method, because `termination_time` is not present in read resources (same problem with devices, but we sanitize it there).

On the other hand, `end_at` is present in the API resource read from `/spot-market-requests/<uuid>`, so it's a better param to use to specify life-span.

I think I should remove the `termin>t<ation_time` from the instance params, in favor of `end_at`, maybe even in this PR.

displague: I'll check with the API team and see if the `termination_time` should be exposed here. I believe it should since the `termination_time` is effective, the device will be terminated, so the API should reveal the field/value.

FWIW, instances that are created via spot market do include this field as an attribute.  I think `termination_time` is being omitted for non spot-market instances and I don't think that should be the case.

t0mk: @displague OK. I was trying to verify if `termination_time` is exported for spot devices, but I can't seem to be able to spin a device from a SMR. I will continue tomorrow.

t0mk: finally I managed to spin a device with SMR which has `end_at` set. The GET to the SMR resource (`/spot-market-requests/289444a5-7f83-439d-90e1-dfc85e843920?include=devices`) looks like:
```json
{
  "id": "289444a5-7f83-439d-90e1-dfc85e843920",
  "created_at": "2021-09-17T13:04:37Z",
  "devices_min": 1,
  "devices_max": 1,
  "max_bid_price": 1.0,
  "end_at": "2021-09-18T21:00:00Z",
  "devices": [
    {
      "termination_time": null,
      "spot_price_max": 1.0,
      "state": "provisioning",
```
So there is a `termination_time`, like you've mentioned, but it's `null`. It's null when the deivce is in `processing` and also when the device is `active`. This si different from on-demand devices where the termination_time is not present at all.

The SMR above was created via the web console, the creating request was sth like (termination_time not present in instance_parameters):

```json
{
  "devices_max": 1,
  "max_bid_price": 1,
  "metro": "sp",
  "instance_parameters": {
    "hostname": "ttt",
    "plan": "c3.medium.x86",
    "operating_system": "ubuntu_16_04",
    "ipxe_script_url": "",
    "userdata": "",
    "quantity": 1,
    "ip_addresses": [
      {
        "address_family": 4,
        "public": false
      },
      {
        "address_family": 4,
        "public": true
      },
      {
        "address_family": 6,
        "public": true
      }
    ]
  },
  "end_at": "2021-09-17T21:00:00Z"
}
```

So I think the `termination_time` is overall unused/useless.

displague: The `termination_time` will only be shown in the SMR devices once the device has been outbid.

It is interesting that `end_dt` does not set a default `termination_time` though.

ocobleseqx:
Due to upcoming deprecation of this repository this pull request has been transferred to https://github.com/equinix/terraform-provider-equinix/pull/210

Please follow the link to the new PR for further discussion

PR #44, opened 2021-03-16
Resource deletes should succeed on 404 and 403 responses

displague:
Fixes #42 
Fixes #43 
Fixes #32

PR #192, opened 2021-09-23
fix 403 and 404 detection

displague:
Revisiting #44 

I received `│ Error: You are not authorized to view this resource` while `terraform destroy`ing a `metal_vlan`.
I also received `Error: Error deleting IP reservation block 4f5fa092-099d-497e-a514-5f25e24b098a: DELETE https://api.equinix.com/metal/v1/ips/4f5fa092-099d-497e-a514-5f25e24b098a: 404 Not found` in https://github.com/equinix/terraform-metal-vsphere/pull/40/checks

While any `metal_vlan` is being deleted an attempt is made to unassign all device ports.  The request returned a 403 because the ports were unavailable because the device was deleted first.

403 and 404 should be handled as success during deletions.  

The problem addressed in this PR is that `ignoreResponseErrors` was returning the first error if it was not the first ignored error.

Additionally, the response objects in the ignoring functions can be safely sent as null. When they are null the error will be examined, which may contain a response.

This error detection has become confusing, in part because the response and error are typically both sent, with the error including the response within it.  It's also no longer obvious to me what value there is in "IsAPIError".

### DIFF
--- a/metal/errors.go
+++ b/metal/errors.go
@@ -41,7 +41,8 @@ func friendlyError(err error) error {
 }
 
 func isForbidden(err error) bool {
-	if r, ok := err.(*packngo.ErrorResponse); ok && r.Response != nil {
+	r, ok := err.(*packngo.ErrorResponse)
+	if ok && r.Response != nil {
 		return r.Response.StatusCode == http.StatusForbidden
 	}
 	if r, ok := err.(*ErrorResponse); ok {
@@ -51,11 +52,11 @@ func isForbidden(err error) bool {
 }
 
 func isNotFound(err error) bool {
-	if r, ok := err.(*packngo.ErrorResponse); ok && r.Response != nil {
-		return r.Response.StatusCode == http.StatusNotFound
-	}
 	if r, ok := err.(*ErrorResponse); ok {
 		return r.StatusCode == http.StatusNotFound && r.IsAPIError
+	}
+	if r, ok := err.(*packngo.ErrorResponse); ok && r.Response != nil {
+		return r.Response.StatusCode == http.StatusNotFound
 	}
 	return false
 }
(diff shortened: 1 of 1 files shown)


displague:
It might be nice to add tests that try to E2E delete a bad UUID.  I haven't seen Terraform tests that set artificial state, but presumably we should be able to create a `metal_vlan` and assert that a `metal_port_vlan_attachment` exists with state that includes the actual vlan id and a fake port `123`. We would then attempt to delete that `metal_port_vlan_attachment`.

t0mk (approved):
Yes, the error processing code looks quite complicated, and quite hard to read. I didn't test this.

I understand what it's trying to do - to check all the possible StatusCode values in a complicated struct.

I also understand that it's hard to clean it, so I give this a green light, but @displague can you maybe write here where each StatusCode gets populated from (I mean from which parts of code of the provider or packngo)? There's

- resp.StatusCode
- err.(*ErrorResponse).StatusCode
- err.(*packngo.ErrorResponse).Response.StatusCode
 
Maybe if we identify the code filling the StatusCode, it will be easier to clean it.

displague:
When any packngo request is sent we get both the HTTP response and an error. 
* https://github.com/packethost/packngo/blob/v0.19.1/packngo.go#L148

Packngo will decorate the API and any HTTP errors as an `ErrorResponse`. In doing so, a reference to the HTTP response is bundled with the error.  
* https://github.com/packethost/packngo/blob/v0.19.1/packngo.go#L71-L81
* https://github.com/packethost/packngo/blob/v0.19.1/packngo.go#L202-L206
* https://github.com/packethost/packngo/blob/v0.19.1/packngo.go#L403-L431

This Terraform provider will also decorate errors (`friendlyError`). If the error was a `packngo.ErrorResponse` a `metal.ErrorResponse` will be returned.  The `ErrorResponse` adds the `StatusCode` (already present in `packngo.ErrorResponse`).

* https://github.com/equinix/terraform-provider-metal/blob/main/metal/errors.go#L14-L41

This Terraform provider attempts to dig deeper into some errors to apply the correct Terraform behavior for acceptable errors (Forbidden NotFound), either of which can indicate a successful delete.

When the Terraform provider digs into errors, the error may not always be a `packngo.ErrorResponse`. The error may be a `metal.ErrorResponse`.  This is because in some call patterns, presumably helpers, `friendlyError()` may have already been applied.

`metal.ErrorResponse` has an `IsAPIError` property which indicates whether or not the error was an API error or an error from a fronting service (loadbalancer, cache).  This is important to capture because a 403/404 from the fronting service could indicate an API delivery problem rather than indicating that EM resources have been removed. 
_Note: This property is used once_.
* https://github.com/equinix/terraform-provider-metal/blob/v0.19.1/metal/errors.go#L24-L30


Thoughts:
* remove `metal.Errors`
  * `packngo.ErrorResponse` implements `Error() string` which provides the same functionality.
* remove `friendlyError`
  * replace `IsAPIError` with an `isAPIError()` helper or an `ignoreErrors(missingMetalAPIHeaders)` pattern
* remove `metal.ErrorResponse`
  * ... ?
* use `multierrors` for the multiple `packngo.ResponseErrors` when offered?
  * create `packngo.Errors` and have `packngo.ResponseErrors` use that? Perhaps `ResponseErrors` should implement a new interface that includes a method to return errors as `[]string`

PR #194, opened 2021-10-05
Datasource metal_plans

t0mk:
This is a new datasource for querying plans, fixing #164 This is just a draft, for discussion.

It's done in the digitalocean style, like 
https://registry.terraform.io/providers/digitalocean/digitalocean/latest/docs/data-sources/images

..with the `filter` blocks. The functionality allowes more than filters, it can also sort, and match by RE.

@displague What do you think? I think that if we want ot accumulate filters, we should do it in a clean and structured way. And here we can take advantage of an existing clean and structured way.

I basically took the filtering logic code from DO provider. It's under the same licences (MPL2), and after brief research, I think there's nothing we should explicitly do in order to have this legit. I also didn't find a way how to properly acknowledge this.


t0mk:
@displague I came back to this. I see that the filtering functionality from #180 clashes with the provider-logic filters.

I think we can't have the API filter and provider logic filter in the same sub-schema. the provider logic filtering acts on the attributes of the datasource, and they might be different from the API resource attributes.

I am not sure if using the API filters have any sense here or in any datasoruce ino the provider. The provider logic filtering is better, it offers
- match by substring, re, exact
- match value from a list

Arguably, using API search can make queries faster for customers that have tens or hundreds of resources.

For the API search, I'd suggest sth like 

```
    api_filter {
        key = "available_in_metros"
        value = "da" // singular
    }
```

displague:
We should add the `reservation_pricing` field to packngo and include that in Terraform. (cc @cprivite)

ocobleseqx:
I want to take this back to improve the acceptance tests as @displague mentioned https://github.com/equinix/terraform-provider-metal/issues/164#issuecomment-884896686

In general, I agree with the above. I would keep client/server filters separate. Adding this extra functionality not available in the API, should not be a problem as long as the difference is well described in the provider documentation. I second the idea of setting  "api_filter" for server-side which explicitly describes that the keys must be the API payload args e.g. "available_in","specs.cpus.count",etc.

Some planned changes:

- replace match_by option "more_than" with "greater_than"
- add match_by options: "less_than_or_equal" and "greater_than_or_equal"
- replace schema field "available_in" with "available_in_facilities"
- add descriptions to all schema fields
- add documentation

Question @displague: should we use "attribute" instead of "key" ? I want to reinforce that it is provider logic and not the API


filter {
  ~~key = "pricing_hour"~~
  attribute = "pricing_hour"
  values = [1]
  match_by = "less_than"
}

displague:
> I think this would be:
```
filter {
  attribute = "facilities" #whatever the field name is "facility"?
  values    = ["da11"]
  match_by  = "in"
}
```

I didn't realize the field name was `available_in`. It makes sense for that to be the attribute in this case.

metal/internal/datalist/filter.go
```diff
@@ -0,0 +1,245 @@
...
+					Required:     true,
+					ValidateFunc: validation.StringInSlice(allowedAttributes, false),
+				},
+				"values": {
+					Type:        schema.TypeList,
+					Description: "The filter values. Filter values are case-sensitive. If you specify multiple values for a filter, the values are joined with an OR by default, and the request returns all results that match any of the specified values",
+					Required:    true,
+					Elem:        &schema.Schema{Type: schema.TypeString},
+				},
+				"all": {
```
displague, line 41: I contemplating suggesting that we drop `all` in favor of `match_by = "contains_all"`, but after looking at the implementation I see how `all` enables necessary boolean operations for other (all?) `match_by` operations.

ocobleseqx: I find it useful the way it is, so you can combine also `all` with other match_by options like `re` (regexp) or `substring`.

```
# example to return all devices containing both "32-core" and "2.5Ghz" strings in the description
data "metal_plans" "example" {
    filter {
        attribute    = "description"
        values        = ["32-core", "2.5Ghz"]
        match_by  = "substring"
        all               = true
    }
}
```

I tried to add a good description to make it easy to understand how it works. If it's not, we can just remove "all" and same behavior can be achieved using several filters instead:

```
# same example to return all devices containing both "32-core" and "2.5Ghz" strings in the description
# using separated filteres over same attribute
data "metal_plans" "example" {
    filter {
        attribute    = "description"
        values        = ["32-core"]
        match_by  = "substring"
    }
    filter {
        attribute    = "description"
        values        = ["2.5Ghz"]
        match_by  = "substring"
    }
}

displague (commented):
This LGTM - we just need a rebase to fix go.mod/go.sum

displague (approved):
Thanks @t0mk and @ocobleseqx!

This PR provides an excellent way to choose the best plans, avoid location-sensitive availability pressures, and a generic filtering feature we can reuse throughout the provider.

PR #200, opened 2021-11-19
Fix for error when deleting device with hardware_reservation_id == next-available

### LINKED ISSUE #199: deleting a device from a hardware reservation fails
Deleting a device created from a hardware reservation fails when `wait_for_reservation_deprovision = "true"` is set.

https://github.com/equinix/terraform-provider-metal/blob/main/metal/resource_metal_device.go#L781

The `deployed_hardware_reservation_id` should be used rather than the `hardware_reservation_id`.

As a fall-back, if `deployed_hardware_reservation_id` is not set, we could consider deleting by `hardware_reservation_id` so long as it is not set to `next-available`. If implemented, this would need to be prepared for a 404.  One reason not to implement this fall-back behavior is that a reservation consumed outside of this TF config could be incorrectly deleted.  Users wanting to clean up a `metal_device` that lacks `deployed_hardware_reservation_id` in this case should import the metal_device (consuming the reservation) first, before attempting to delete it.

t0mk:
This PR fixes device deletion when provider waits for deprovision of hw reservation AND the hardware_reservation_id was specified as `next-available`.

fixes #199 

### DIFF
--- a/metal/resource_metal_device.go
+++ b/metal/resource_metal_device.go
@@ -778,7 +778,7 @@ func resourceMetalDeviceDelete(d *schema.ResourceData, meta interface{}) error {
 		return friendlyError(err)
 	}
 
-	resId, resIdOk := d.GetOk("hardware_reservation_id")
+	resId, resIdOk := d.GetOk("deployed_hardware_reservation_id")
 	if resIdOk {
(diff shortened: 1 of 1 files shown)


t0mk:
Let me just test this.

t0mk:
OK, it seems it's working.

PR #204, opened 2021-11-25
Fix project payment method

### LINKED ISSUE #201: metal_project payment_method_id UUID validation prevents refresh
It is possible for the payment_method to not be included in the project API response, even when `payment_method` is included. This is due to some billing relationship or token access to billing information.

When payment info is not included in the API response, the `PaymentMethod` object has zero values, including the Href and ID. This results in UUID validation by Terraform.

During refresh:
```
invalid value for payment_method_id (must be a valid UUID)
```

Packngo should reflect that Project PaymentMethod can be null. (it may be omitted by the API)
Terraform validation should allow this field to be empty.
https://github.com/equinix/terraform-provider-metal/blob/main/metal/resource_metal_project.go#L59

### Terraform Configuration Files
```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file.
```

### Debug Output
Please provider a link to a GitHub Gist containing the complete debug output: https://www.terraform.io/docs/internals/debugging.html. Please do NOT paste the debug output in the issue; just paste a link to the Gist. Note: the debug log will contain your API key, please review the log and delete sensitive secrets before submitting.

### Panic Output
If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the `crash.log`.

### Expected Behavior
What should have happened?

### Actual Behavior
What actually happened?

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

### Important Factoids
Are there anything atypical about your accounts that we should know?

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:
- #1234

displague:
Fixes #201


metal/resource_metal_project_test.go
```diff
@@ -88,6 +88,11 @@ func TestAccMetalProject_BGPBasic(t *testing.T) {
 						"2SFsdfsg43"),
 				),
 			},
+			{
```
t0mk, line 91: If you add the import check to the project tests, why not adding it to the `TestAccMetalProject_Basic` too?

metal/resource_metal_project.go
```diff
@@ -56,7 +56,10 @@ func resourceMetalProject() *schema.Resource {
 				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
 					return strings.EqualFold(strings.Trim(old, `"`), strings.Trim(new, `"`))
 				},
-				ValidateFunc: validation.StringMatch(uuidRE, "must be a valid UUID"),
+				ValidateFunc: validation.Any(
+					validation.StringMatch(uuidRE, "must be a valid UUID"),
+					validation.StringIsEmpty,
+				),
```
t0mk, line 62: Looks good. I can't say if it fixes the problem though, as haven't been able to invoke/replicate the error from 
#201

t0mk (changes requested):

metal/resource_metal_project_test.go
displague: I did at first and then noticed `TestAccMetalProject_importBasic` which does the same.

PR #207, opened 2022-01-03
Fabric service tokens in metal_connection

t0mk:
This PR adds Fabric service token support to metal_connection resource and datasource.

I also changed attribute validation in metal_connection, according to my understanding of what has changed (in `resourceMetalConnectionCreate`)

There are 2 things to do before this can be merged:

- This PR is blocked by https://github.com/packethost/packngo/issues/317, because right now it's not possible to read project ID for a created connection. Once the `project` attribute is added to Connection resoruce in EM API, it should be added to packngo, and then the read of `project_id` in `resourceMetalConnectionRead` in `metal/resource_metal_connection.go` should be updated.

- Update go mod dependency when the fabric service token support is released in packngo.

### DIFF
--- a/docs/data-sources/connection.md
+++ b/docs/data-sources/connection.md
@@ -23,19 +23,26 @@ data "metal_connection" "example" {
 
 ## Attributes Reference
 
-* `description` - Description of the connection resource
 * `name` - Name of the connection resource
-* `tags` - String list of tags
-* `facility` - Slug of a facility to which the connection belongs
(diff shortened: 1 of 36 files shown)


t0mk:
This is blocked by https://github.com/packethost/packngo/issues/317

displague (approved):
Thanks for precisely capturing this new feature of the EM API and for the detail in the implementation!

PR #209, opened 2022-01-14
wait for IP Reservations to reach the desired state

### LINKED ISSUE #135: metal_reserved_ip_block should be returned when 'created'
A module can't function with a cidr_notation value of `/0`.  Terraform resources are generally expected to return when the resource is ready to use. 

The Terraform provider should adopt a behavior of waiting for the`metal_reserved_ip_block` to have a state of `created` before releasing the resource.  In some cases, a user may be ok with this resource being released early, so a behavior flag could be used to toggle that.

_Originally posted by @displague in https://github.com/equinix/terraform-metal-anthos-on-baremetal/issues/74#issuecomment-865215226_

displague:
This PR adds various features to `metal_reserved_ip_block`:
* Adds `custom_data` (a string of JSON. This enables the use of arbitrary scalar, list, and map types, as supported by the API. This is defaulted to the API default of `"{}"`.
* Allows for `tags`, `description`, and `custom_data` to be updated
* Adds `wait_for_state` (defaults to `created`)
* The default behavior is now to wait for the IP Address reservation to be usable (guaranteeing an IP Address) before exiting the create call. A number of users have been snagged by this resource releasing after creating the reservation request without having an IP address fully reserved.  Those users will now get a 10m timeout (default, configurable with `timeouts`) unless they set `wait_for_state="pending"`

Fixes #135

### DIFF
--- a/metal/resource_metal_reserved_ip_block.go
+++ b/metal/resource_metal_reserved_ip_block.go
@@ -62,11 +62,6 @@ func metalIPComputedFields() map[string]*schema.Schema {
 
 func metalIPResourceComputedFields() map[string]*schema.Schema {
 	s := metalIPComputedFields()
-	s["address_family"] = &schema.Schema{
-		Type:        schema.TypeInt,
-		Computed:    true,
-		Description: "Address family as integer (4 or 6)",
-	}
 	s["public"] = &schema.Schema{
 		Type:        schema.TypeBool,
 		Computed:    true,
--- a/metal/resource_metal_reserved_ip_block.go
+++ b/metal/resource_metal_reserved_ip_block.go
@@ -59,6 +59,8 @@ func metalIPComputedFields() map[string]*schema.Schema {
 			Type:        schema.TypeString,
 			Description: "Wait for the IP reservation block to reach a desired state on resource creation. One of: `pending`, `created`. The `created` state is default and recommended if the addresses are needed within the configuration. An error will be returned if a timeout or the `denied` state is encountered.",
 			Default:     packngo.IPReservationStateCreated,
+			Optional:    true,
+			ForceNew:    false,
 			ValidateDiagFunc: validation.ToDiagFunc(
 				validation.StringInSlice([]string{
 					string(packngo.IPReservationStateCreated),
--- a/metal/resource_metal_reserved_ip_block.go
+++ b/metal/resource_metal_reserved_ip_block.go
@@ -1,6 +1,7 @@
 package metal
 
 import (
+	"encoding/json"
 	"fmt"
 	"log"
 	"path"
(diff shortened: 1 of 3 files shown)


metal/resource_metal_reserved_ip_block.go
```diff
@@ -49,6 +55,19 @@ func metalIPComputedFields() map[string]*schema.Schema {
 			Type:     schema.TypeBool,
 			Computed: true,
 		},
+		"wait_for_state": {
+			Type:        schema.TypeString,
+			Description: "Wait for the IP reservation block to reach a desired state on resource creation. One of: `pending`, `created`. The `created` state is default and recommended if the addresses are needed within the configuration. An error will be returned if a timeout or the `denied` state is encountered.",
+			Default:     packngo.IPReservationStateCreated,
+			Optional:    true,
+			ForceNew:    false,
```
ocobleseqx, line 63: ForceNew=false requires the resource to have Update function defined. You can just define a dummy Update function returning Read() or set ForceNew=true like other parameters that can be updated in the UI portal have -->`Description` / `tags`

Even not required for this one-time parameter, the update method should be included in the packetngo client to match the API https://developer.equinix.com/catalog/metalv1#operation/updateIPAddress and then modify behavior here for Description/tags

metal/resource_metal_reserved_ip_block.go
```diff
@@ -49,6 +55,19 @@ func metalIPComputedFields() map[string]*schema.Schema {
 			Type:     schema.TypeBool,
 			Computed: true,
 		},
+		"wait_for_state": {
```
ocobleseqx, line 58: Makes more sense to define `wait_for_state` directly in resourceMetalReservedIPBlock() since it's not computed. 

(This should be managed in a different issue/pr) Even more, I'm not sure why there is a distinction for some computed params defined in two different function metalIPResourceComputedFields() and metalIPComputedFields(). Having a single function would simplify code maintaining  and avoid errors, for example there I found `address_family` is defined twice. 
https://github.com/equinix/terraform-provider-metal/blob/3834e17fb84992b49c8b758faf358bd6ff852867/metal/resource_metal_reserved_ip_block.go#L26

https://github.com/equinix/terraform-provider-metal/blob/3834e17fb84992b49c8b758faf358bd6ff852867/metal/resource_metal_reserved_ip_block.go#L76

ocobleseqx (changes requested):
I left a couple of comments that should be addressed before merge.

```
with the provider itself, and not a user issue. Please report
│ this bug:
│ 
│       * resource metal_reserved_ip_block: No Update defined, must set ForceNew on: []string{"wait_for_state"}
```

metal/resource_metal_reserved_ip_block.go
displague: We are tracking this issue https://github.com/packethost/packngo/issues/318. When the [PR that addresses this](https://github.com/packethost/packngo/pull/319) has been merged, I'll update this PR to include the Update functionality.

metal/resource_metal_reserved_ip_block.go
displague: I think the distinction owes to packngo's definition of `IPAddressReservation` and `IPAddressCommon`. And this owes to the API spec and (more importantly API responses) where reservations (user requested IP blocks) have different properties than non-reservations (dynamically or statically assigned addresses that the user can reclaim/reuse).

displague: That said, if there are overlapping aspects of the code that can be refactored, we should look into that.

displague: I've removed the duplicate definition.

This helper shares schema definitions between metal_ip_attachment, data.metal_precreated_ip_block, metal_resource_ip_block. This is something we can revisit later.

metal/resource_metal_reserved_ip_block.go
displague: Update functionality has been added.

ocobleseqx (approved):
Looks good. Nothing to add from my side

PR #211, opened 2022-02-07
wait_for_reservation_deprovision should succeed if a different instance id is found

### LINKED ISSUE #208: metal_device wait_for_deprovision can be raced with a new instance claiming the hardware reservation
It is possible (seen by real users) for a `wait_for_deprovision` enabled `metal_device` to be destroyed and then reclaimed by an external tool BEFORE the `wait_for_deprovision` destroy loop detects that the device was provisionable.  The machine is then unprovisionable and the full 1hr timeout must be awaited.

Rather than waiting for `provisionable=true`, the RefreshStateFunc could wait for the instance_id of the reservation to become empty and for provisionable flag to be true.  The known instance_id would be used in the `Pending` slice of strings that this helper function takes, with a `Target` of `""`.  If an unknown instance_id is returned by the function, Terraform would fail because of the unexpected value.

displague:
Fixes #208 

When a Device Instance uses a Hardware Reservation ID, the `wait_for_reservation_deprovision` parameter can be set to instruct the Terraform provider to not release the `metal_device` delete process until the hardware reservation has reached the `provisionable` state.  The hardware reservation API resource will be polled on a refresh cycle until it is found in `provisionable=true` state.

In some user-scenarios, the hardware reservation may be reassigned by another user or other software while this Terraform provider is polling.  When this occurs, the resource may never reach `provisionable=true`.

From a Terraform user perspective, the best behavior the provider can ascribe to this scenario is to stop polling when `provisionable=false` is detected but the reservation's `instance.ID` no longer matches the instance ID that was being deleted.  The deprovision cycle has been completed, so Terraform should move on.  If the user intended to reuse that reservation in their own configuration, the recreation phase will return an error that the reservation is already in use.

This PR adds this behavior and mock tests to verify the behavior.
A log message will `WARN` if a reprovisioned reservation has been detected.

### DIFF
--- a/metal/helpers_device.go
+++ b/metal/helpers_device.go
@@ -2,6 +2,7 @@ package metal
 
 import (
 	"fmt"
+	"log"
 	"strings"
 	"sync"
 	"time"
@@ -128,13 +129,10 @@ func hwReservationStateRefreshFunc(client *packngo.Client, reservationId, instan
 		case r != nil && r.Provisionable:
 			state = provisionable
 		case r != nil && r.Device != nil && (r.Device.ID != "" && r.Device.ID != instanceId):
+			log.Printf("[WARN] Equinix Metal device instance %s (reservation %s) was reprovisioned to a another instance (%s)", instanceId, reservationId, r.Device.ID)
 			state = reprovisioned
 		}
-		p := false
-		if r != nil {
-			p = r.Provisionable
-		}
-		fmt.Println("provisionable", p, "state", state, "err", err, "reservation", r)
+
 		return r, state, err
 	}
 }
--- a/metal/helpers_device.go
+++ b/metal/helpers_device.go
@@ -110,21 +110,25 @@ func getPorts(ps []packngo.Port) []map[string]interface{} {
 	return ret
 }
 
-func waitUntilReservationProvisionable(id string, meta interface{}) error {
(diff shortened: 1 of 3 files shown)


displague (commented):
I'm adding local mocking tests, `Test_waitUntilReservationProvisionable`. This should be ready tomorrow.

metal/helpers_device.go
```diff
@@ -110,25 +110,43 @@ func getPorts(ps []packngo.Port) []map[string]interface{} {
...
+		r, _, err := client.HardwareReservations.Get(reservationId, nil)
+		state := deprovisioning
+		switch {
+		case err != nil:
+			err = friendlyError(err)
+			state = errstate
+		case r != nil && r.Provisionable:
+			state = provisionable
+		case r != nil && r.Device != nil && (r.Device.ID != "" && r.Device.ID != instanceId):
+			state = reprovisioned
```
displague, line 131: We should log when a hardware reservation was reprovisioned to another instance while awaiting deprovision.  This is a good behavior that will allow Terraform to exit at the right time while waiting for deprovision, but it will be confusing to users that expected a full deprovision (likely awaiting a reprovision of their own).

ocobleseqx (approved):
lgtm 👍🏼

PR #217, opened 2022-03-14
include device filter in hwReservationStateRefreshFunc

ocobleseqx:
This PR aims to fix https://github.com/equinix/terraform-provider-metal/issues/208 and is related to https://github.com/equinix/terraform-provider-metal/pull/211

As described in the issue, get function was taking a nil packngo.GetOptions and therefore returning a nil Device
required to check it Device ID has changed.

https://github.com/equinix/terraform-provider-metal/blob/edc45fee397d331d03bdd3722b52641cd4354936/metal/helpers_device.go#L123

It has been replaced:

`r, _, err := client.HardwareReservations.Get(reservationId, &packngo.GetOptions{Includes: []string{"device"}})`

### DIFF
--- a/metal/helpers_device.go
+++ b/metal/helpers_device.go
@@ -120,7 +120,7 @@ const (
 
 func hwReservationStateRefreshFunc(client *packngo.Client, reservationId, instanceId string) resource.StateRefreshFunc {
 	return func() (interface{}, string, error) {
-		r, _, err := client.HardwareReservations.Get(reservationId, nil)
+		r, _, err := client.HardwareReservations.Get(reservationId, &packngo.GetOptions{Includes: []string{"device"}})
 		state := deprovisioning
 		switch {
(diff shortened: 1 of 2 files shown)


displague (approved):
This is excellent. The updated mock now dynamically includes the device data that was missing in the real-world use scenario and the change to include the device details fixes the problem we were seeing.

In a future PR we should add the additional logging that was mentioned in #208.

I've created https://github.com/equinix/terraform-provider-metal/issues/218 to track this.

displague (approved):
This is excellent. The updated mock now dynamically includes the device data that was missing in the real-world use scenario and the change to include the device details fixes the problem we were seeing.

In a future PR we should add the additional logging that was mentioned in #208.

I've created https://github.com/equinix/terraform-provider-metal/issues/218 to track this.

PR #219, opened 2022-03-25
consider metal_device delete timeout settings when awaiting hardware deprovision

displague:
Fixes 218
Fixes 216


displague:
Possible related conversation here https://github.com/hashicorp/terraform-plugin-sdk/issues/675

Issue #222, opened 2022-05-09
r/equinix_metal_connection produces non-empty plan after apply if type is dedicated


ocobleseqx:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

https://github.com/equinix/terraform-provider-equinix/blob/314dead5ac3ec831beb176bc98d1e5c84a80f242/equinix/resource_metal_connection.go#L156

Since project_id is required, the specified value should be returned always, otherwise it forces recreation of the resource.

### Terraform Version
Run `terraform -v` to show the version. If you are not running the latest version of Terraform, please upgrade because your issue may have already been fixed.

Terraform v1.0.11
on linux_amd64

Provider version: v3.2.2

### Affected Resource(s)
Please list the resources as a list, for example:
- resource_metal_connection

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
        resource "equinix_metal_project" "test" {
            name = "my-project"
        }

        resource "equinix_metal_connection" "test" {
            name = "my-conn"
            project_id = equinix_metal_project.test.id
            metro = "sv"
            redundancy = "redundant"
            type = "dedicated"
            mode = "tunnel"
            speed = "50Mbps"
        }
```

### Actual Behavior

What actually happened?
![image](https://user-images.githubusercontent.com/68897552/167365744-b3f47a8c-8c51-4b6c-9895-14987c890688.png)

ocobleseqx:
It also produces an error in datasource equinix_metal_connection if connection was created in the portal without specifying speed value

```
╷
│ Error: 0 is not allowed speed value. Allowed values: [50000000 200000000 500000000 1000000000 2000000000 5000000000 10000000000]
│ 
│   with module.vlan["ch"].data.metal_connection.fabric_connection,
│   on .terraform/modules/vlan/data.tf line 1, in data "metal_connection" "fabric_connection":
│    1: data "metal_connection" "fabric_connection" {
│ 
```

PR #223, opened 2022-05-11
Fix resource metal connection

ocobleseqx:
- Always return project_id specified in the schema (to cover dedicated connection without project), and later for shared connections check project assigned to the virtual circuit.
- Allow retrieve existing connections created through the portal with speed value 0.
- Sort ports by role , asserting the API always returns primary for len of 1 responses (to replicate current provider equinix behavoir).

fix [#222 ](https://github.com/equinix/terraform-provider-metal/issues/222)


metal/resource_metal_connection.go
```diff
@@ -104,8 +104,8 @@ func resourceMetalConnection() *schema.Resource {
 			},
 			"project_id": {
 				Type:        schema.TypeString,
-				Required:    true,
-				Description: "ID of the project where the connection is scoped to, only used for type == \"shared\"",
+				Optional:    true,
+				Description: "ID of the project where the connection is scoped to. Required with type \"shared\"",
 				ForceNew:    true,
```
displague, line 109: Is this going to create an issue for users with dedicated connections that have a project id specified today? (Is that possible or was that only introduced within the alpha builds?)

Perhaps we should use custom validation to allow users to move from project to the same org without triggering a forcenew.  Or perhaps we should include a changelog note on how to convert to organization without losing the connection.

ocobleseqx: It was introduces in alpha. Previously (3.2.2), "organization_id" was Required and "project_id" Optional and connection was always created through the org endpoint but validating (internally) that project was specified when type was ConnectionShared. Now, main difference is that if organization is not specified it will take the org to which the project belongs. So if anyone has created a dedicated connection with any alpha release (with org computed and with project required) now they can continue without org and with project because it first checks "organization_id" is set, otherwise it checks "project_id" is set to retrieve the organization id.

https://github.com/equinix/terraform-provider-metal/pull/223/files/207d44fa81df1ee0bb861dcdfd45e4527ada5914#diff-fb65e05d73d1a84ee2934a9c6033312470724448442603e716a9f6fada29c4eaR260-R266

displague (approved):

PR #224, opened 2022-05-25
backport VRF support from equinix/equinix provider

displague:
This is a backport of equinix/terraform-provider-equinix#129

Mostly derived from
https://patch-diff.githubusercontent.com/raw/equinix/terraform-provider-equinix/pull/129.patch
with some automated changes to the patch and manually cherry-picked changes from the rejects.

```
# vim commands applied against the patch
:%s_equinix/_metal/
:%s/equinix_//
:%s/data_source_metal/datasource_metal
:%s/_acc_test.go/_test.go
:%s/package equinix/package metal
:%s#docs/resources/metal_#docs/resources/
:%s#docs/data-sources/metal_#docs/data-sources/
```

Patch applied with:

```
patch -p1 --input=129.patch --fuzz=16
```

Rejects reviewed with:
```
find . -name \*rej  -exec less {} \;
```

Notably, rejects that dealt with test locations and plans were ignored for this PR.
Some helper tools for set/list conversion and sweeper test detection where copied from equinix/equinix.

### DIFF
--- a/metal/resource_metal_reserved_ip_block_test.go
+++ b/metal/resource_metal_reserved_ip_block_test.go
@@ -35,7 +35,7 @@ resource "metal_project" "foobar" {
 
 resource "metal_reserved_ip_block" "test" {
 	project_id  = metal_project.foobar.id
-	facility    = "ny5"
+	facility    = "sv"
 	type        = "public_ipv4"
 	quantity    = 2
 	tags        = ["Tag1", "Tag2"]
@@ -101,7 +101,7 @@ func TestAccMetalReservedIPBlock_Public(t *testing.T) {
 				Config: testAccCheckMetalReservedIPBlockConfig_Public(rs),
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr(
-						"metal_reserved_ip_block.test", "facility", "ewr1"),
+						"metal_reserved_ip_block.test", "facility", "sv"),
 					resource.TestCheckResourceAttr(
 						"metal_reserved_ip_block.test", "type", "public_ipv4"),
 					resource.TestCheckResourceAttr(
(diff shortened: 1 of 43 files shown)


metal/resource_metal_reserved_ip_block_test.go
```diff
@@ -35,7 +35,7 @@ resource "metal_project" "foobar" {
 
 resource "metal_reserved_ip_block" "test" {
 	project_id  = metal_project.foobar.id
-	facility    = "ny5"
+	facility    = "sv"
```
displague, line 38: This should be `metro =` (and line 104 should be metro too). Similar changes appear elsewhere.

metal/resource_metal_reserved_ip_block_test.go
```diff
@@ -230,15 +230,15 @@ resource "metal_project" "foobar" {
 
 resource "metal_reserved_ip_block" "test" {
 	project_id  = metal_project.foobar.id
-	facility    = "ny5"
+	facility    = "ewr1"
```
displague, line 233: We would want to go from ewr1 to ny5, not reverse.

metal/resource_metal_spot_market_request_test.go
```diff
@@ -17,8 +17,8 @@ resource "metal_project" "test" {
 }
 
 data "metal_spot_market_price" "test" {
-  facility = "sv15"
-  plan     = "c3.small.x86"
+  facility = "ewr1"
```
displague, line 20: sv15 would be preferable here. I think c3.small would be preferable as well.

metal/resource_metal_reserved_ip_block_test.go
ocobleseqx: metros and facilities have been replaces with da/da11

metal/resource_metal_spot_market_request_test.go
ocobleseqx: replaced with c3.medium

PR #229, opened 2022-06-13
check service_token expiresAt attribute is null

ocobleseqx:
Added validation to check expires_at is not null before converting Timestamp to string

Bug fix https://github.com/equinix/terraform-provider-metal/issues/226

Packngo version update required to v0.25.0

Sync from https://github.com/equinix/terraform-provider-equinix/pull/148

### DIFF
--- a/docs/data-sources/gateway.md
+++ b/docs/data-sources/gateway.md
@@ -9,6 +9,8 @@ description: |-
 
 Use this datasource to retrieve Metal Gateway resources in Equinix Metal.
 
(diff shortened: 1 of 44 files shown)


ocobleseqx:
Upgrade of packngo version requires this PR https://github.com/equinix/terraform-provider-metal/pull/224 is merged before

displague:
The VRF backport was squashed out of this which is fine. We can reintroduce that in another PR.

Issue #230, opened 2022-06-13
r/metal_organization add required field address


ocobleseqx:
Sync with terraform equinix provider required, already implemented in https://github.com/equinix/terraform-provider-equinix/issues/134

PR #231, opened 2022-06-14
field address added to rs/ds metal_organization

ocobleseqx:
This PR adds field address (mandatory now to create an organization) in both resource and data source metal_organization.

Sync with https://github.com/equinix/terraform-provider-equinix/pull/137
- Allow retrieve previous organization without address.
- Test updated.
- Documentation updated.

fix https://github.com/equinix/terraform-provider-metal/issues/230

### DIFF
--- a/metal/datasource_metal_organization.go
+++ b/metal/datasource_metal_organization.go
@@ -143,6 +143,7 @@ func dataSourceMetalOrganizationRead(d *schema.ResourceData, meta interface{}) e
 		projectIds = append(projectIds, path.Base(p.URL))
 	}
 
+	d.SetId(org.ID)
 	return setMap(d, map[string]interface{}{
 		"organization_id": org.ID,
 		"name":            org.Name,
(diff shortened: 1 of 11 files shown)


metal/datasource_metal_organization.go
```diff
@@ -55,6 +53,46 @@ func dataSourceMetalOrganization() *schema.Resource {
 				Computed:    true,
 				Elem:        &schema.Schema{Type: schema.TypeString},
 			},
+			"address": {
+				Type:        schema.TypeList,
+				Description: "Business' address",
+				Computed:    true,
+				Optional:    true,
```
displague, line 60: `Optional: true` here (and in createOrganizationAddressDataSourceSchema) implies that address fields can be used as search filters. For the datasource, Optional should be false. 

For the resource, I think the address fields should not be computed, they should be whatever the user has defined in Terraform.

ocobleseqx: Removed Optional from address field in DS schema. Same for address sub-schema fields. Resource was fine without compute, nevertheless I have removed those occurrences of required=false and optional=false to make it cleaner

PR #234, opened 2022-06-15
Fix test sweepers for metal projects

### LINKED ISSUE #233: test sweepers is leaving some projects undeleted
image

ocobleseqx:
To remove projects all other resources that belong to that project must be removed first. There was some old projects not removed with VLANs on them

Sweeper for VLANs uses VLAN description to identify which must be deleted. To make it work ,all them must have same prefix, but some had no description or they were using a random strings. Same happens with virtual circuits

In this PR all descriptions of VLAN resources, and names of virtual circuits created in tests have been modified to have `tfacc` prefix

fix #233

### DIFF
--- a/metal/datasource_metal_device_test.go
+++ b/metal/datasource_metal_device_test.go
@@ -40,7 +40,7 @@ func TestAccDataSourceMetalDevice_Basic(t *testing.T) {
 func testDataSourceMetalDeviceConfig_Basic(projSuffix string) string {
 	return fmt.Sprintf(`
 resource "metal_project" "test" {
-    name = "tfacc-project-%s"
+    name = "tfacc-pro-device-%s"
 }
 
 resource "metal_device" "test" {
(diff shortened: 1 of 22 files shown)


displague (approved):
nit: I understand shortening `-project-` to `-pro-`, but I question the addition of `-pro-` throughout various resource names that did not have this.

Otherwise, LGTM. The sweeper appears to be fixed in this PR and I verified with a manual script.

ocobleseqx:
> nit: I understand shortening `-project-` to `-pro-`, but I question the addition of `-pro-` throughout various resource names that did not have this.
> 
> Otherwise, LGTM. The sweeper appears to be fixed in this PR and I verified with a manual script.

Project' sweeper still only looks for `tfacc-`, but I intend to use same prefix for all projects `tfacc-pro-` as a naming convention which can help others to identify it as a rule, or in the future for an analysis of resources created or even define a more granular sweeper. The next part of the name must refer to the resource being tested, i.e. if test is for resource Virtual Circuit then the name of project created in that plan must be "tfacc-pro-vc-", however I just saw that not all of them were updated like this and I guess that's where the confusion comes from.

PR #239, opened 2022-07-05
add ds plans in all tests with a metal_device  to check available hardware

ocobleseqx:


displague (approved):
This will really help out our test success rate. Thanks!

